PR #127, opened 2020-02-26
[bugfix] fix Stream.onTerminate causing illegal state exception.

Laimiux:
It was not easy to replicate, but the exception is as follows:
java.lang.IllegalStateException: Callback class com.instacart.formula.fragment.FragmentFlowStore$evaluate$$inlined$eventCallback$1 is already defined. Are you calling it in a loop or reusing a method? You can wrap the call with FormulaContext.key

### DIFF
--- a/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
+++ b/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
@@ -29,10 +29,11 @@ internal class FormulaManagerImpl<Input, State, RenderModel>(
 
     private var state: State = state
     private var onTransition: ((Effects?, isValid: Boolean) -> Unit)? = null
+    private var pendingRemoval: MutableList<FormulaManager<*, *, *>>? = null
 
     private fun handleTransition(transition: Transition<State>, wasChildInvalidated: Boolean) {
         if (terminated) {
-            // We only pass messages up.
+            // State transitions are ignored, only side effects are passed up to be executed.
             onTransition?.invoke(transition.effects, true)
             return
         }
@@ -68,7 +69,6 @@ internal class FormulaManagerImpl<Input, State, RenderModel>(
         transitionId: Long
     ): Evaluation<RenderModel> {
         // TODO: assert main thread.
-
         val lastFrame = frame
         if (lastFrame != null && lastFrame.isValid(input)) {
             updateTransitionNumber(transitionId)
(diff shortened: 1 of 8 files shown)


formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -159,29 +173,33 @@ internal class FormulaManagerImpl<Input, State, RenderModel>(
...
-
         return manager.evaluate(formula, input, processingPass).renderModel
     }
 
     override fun markAsTerminated() {
         terminated = true
         frame?.transitionCallbackWrapper?.terminated = true
-
         callbacks.disableAll()
-        updateManager.terminate()
```
Laimiux: `markAsTerminated` is called before we are ready to perform side-effects.  We need to delay `UpdateManager.terminate` until we are ready to handle transitions.

carrotkite:
### JaCoCO Code Coverage 75.88% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/internal/FormulaManagerImpl` | 79% | 0% | :white_check_mark: |
| `com/instacart/formula/test/TestFormulaManager` | 0% | 0% | :white_check_mark: |


</p>

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -159,29 +176,33 @@ internal class FormulaManagerImpl<Input, State, RenderModel>(
         @Suppress("UNCHECKED_CAST")
         val manager = children
             .findOrInit(key) {
-                childManagerFactory.createChildManager(formula, input, transitionLock)
+                childManagerFactory.createChildManager(formula, input, transitionLock).apply {
+                    setTransitionListener { message, isValid ->
```
Laimiux: We only need to set the listener once when we initialize the child manager.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -91,15 +91,28 @@ internal class FormulaManagerImpl<Input, State, RenderModel>(
...
+        pendingRemoval = null
+        local?.forEach { it.performTerminationSideEffects() }
+        if (transitionLock.hasTransitioned(currentTransition)) {
+            return true
+        }
+
+        return children.any { it.value.value.terminateDetachedChildren(currentTransition) }
+    }
+
+    // TODO: should probably terminate children streams, then self.
```
Laimiux: This is a TODO for a later time.

Laimiux:
It was pretty hard to replicate this issue, but I was finally able to do it. It only happens if parent formula produces a new state by `onInputChanged()` and at the same time, a nested child fires a termination event.

PR #128, opened 2020-02-28
Avoid exhausting the stack when doing transitions.

Laimiux:
Updating formula runtime to use a while loop to iterate over side-effects instead of recursion based behavior. The stack still overflows at 500 nested child formulas, however for a different reason than what this PR fixes. Without this change, it would overflow at 100 child formulas.


carrotkite:
### JaCoCO Code Coverage 74.98% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/FormulaRuntime` | 50% | 0% | :white_check_mark: |


</p>

PR #133, opened 2020-03-31
Use startedActivity before posting update

erikwolfe02:
Use `startedActivity` instead of `currentActivity` with `update`

### DIFF
--- a/formula-android/src/main/java/com/instacart/formula/integration/ActivityHolder.kt
+++ b/formula-android/src/main/java/com/instacart/formula/integration/ActivityHolder.kt
@@ -45,7 +45,9 @@ class ActivityHolder<Activity : FragmentActivity> {
     }
 
     fun detachActivity(activity: Activity) {
-        this.activity = null
+        if (this.activity == activity) {
+            this.activity = null
+        }
         lifecycleEventRelay.accept(Unit)
     }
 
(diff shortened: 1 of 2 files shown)


formula-android/src/main/java/com/instacart/formula/integration/ActivityHolder.kt
```diff
@@ -45,9 +45,7 @@ class ActivityHolder<Activity : FragmentActivity> {
     }
 
     fun detachActivity(activity: Activity) {
-        if (this.activity == activity) {
-            this.activity = null
-        }
+        this.activity = null
```
erikwolfe02: This check is not needed since `StoreManager` already confirms they're the same

Laimiux: I think I had this check in case the new activity is created before old one is destroyed.

carrotkite:
### JaCoCO Code Coverage 76.66% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/integration/ActivityHolder` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/integration/ActivityStoreContext` | 33% | 0% | :white_check_mark: |


</p>

formula-android/src/main/java/com/instacart/formula/integration/ActivityHolder.kt
erikwolfe02: Ah interesting. I'll put back just in case.

PR #138, opened 2020-04-18
[sample] creating a custom stream implementation.

Laimiux:
Adding an example that shows how to create a custom stream implementation. It also shows how to hide split it into an interface and the implementation.

Important classes:
- Stream example classes -> `NetworkStateStream.kt` and `NetworkStateStreamImpl.kt`
- Formula using the stream -> `NetworkStateFormula.kt`
- Text example which uses a fake stream implementation -> `NetworkStateFormulaTest.kt`


carrotkite:
### JaCoCO Code Coverage 77.11% :white_check_mark:
|:---|:---:|:---:|:---:|


</p>

samples/custom-network-state-stream/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+
+androidExtensions {
+    experimental = true
+    features = ['parcelize']
+}
+
+android {
+    compileSdkVersion 28
+    defaultConfig {
+        applicationId "com.instacart.formula.stopwatch"
```
Jawnnypoo, line 13: Should this change? Probably doesn't matter, but in case someone wanted to install all the samples at once.

Laimiux: Oops, missed during copy-paste.

PR #140, opened 2020-05-14
[android] Rely on actual backstack entries to detect fragments being popped

mhartvick:
This modification keeps track of the actual backstack in `FragmentFlowRenderView` instead of a count. It also keeps a list of fragments being popped to avoid `onLifecycleEvent()` triggered for a wrong fragment, in case of the framework not calling `onFragmentViewDestroyed()` in order.

### DIFF
--- a/formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
+++ b/formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
@@ -176,12 +176,16 @@ internal class FragmentFlowRenderView(
         val backStackEntryCount = backStackEntries.size
         val newBackStackEntryCount = activity.supportFragmentManager.backStackEntryCount
         if (backStackEntryCount > newBackStackEntryCount) {
-            val poppedFragmentName = backStackEntries.removeAt(backStackEntryCount - 1).name
-            if (poppedFragmentName != null && visibleFragments.find { it.tag == poppedFragmentName } != null) {
-                awaitingRemoval.add(poppedFragmentName)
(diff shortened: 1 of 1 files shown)


carrotkite:
### JaCoCO Code Coverage 77.06% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/integration/FragmentFlowRenderView` | 56% | 0% | :white_check_mark: |


</p>

formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
```diff
@@ -169,10 +173,15 @@ internal class FragmentFlowRenderView(
     }
 
     private fun recordBackstackChange() {
-        val newBackstackSize = activity.supportFragmentManager.backStackEntryCount
-        if (backstackEntries > newBackstackSize) {
-            backstackPopped = true
+        val backStackEntryCount = backStackEntries.size
+        val newBackStackEntryCount = activity.supportFragmentManager.backStackEntryCount
+        if (backStackEntryCount > newBackStackEntryCount) {
+            val poppedFragmentName = backStackEntries.removeAt(backStackEntryCount - 1).name
```
Laimiux, line 179: I wonder if we can assume that only `1` entry can be removed at the time.

mhartvick: I actually probably shouldn't assume that (with transactions and other things, I wouldn't be surprised if it's not always the case + seems like fragment is slightly bug prone :trollface:, so we better protect ourselves)

PR #143, opened 2020-06-01
Fix crash when multiple fragments get removed in same "backstackChanged" callback

mhartvick:
Looks like https://github.com/instacart/formula/blob/master/formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt#L180 has been causing a couple of `IndexOutOfBoundException` crashes. 
~Might be caused by some race (and `OnBackStackChangedListener` called from multiple threads)~
Most likely caused by the loop removing elements in "order" and list size decreasing while looping

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,7 +105,7 @@ commands:
       - run:
           name: Assemble Todo App
           command: |
-            ./gradlew :samples:todoapp:assembleRelease
+            ./gradlew :samples:todoapp:assembleRelease --no-daemon
       - run:
           name: Assemble Custom Stream App
           command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -105,7 +105,7 @@ commands:
       - run:
           name: Assemble Todo App
           command: |
-            ./gradlew :samples:todoapp:assembleRelease
```
Laimiux, line 108: I would change all of the `./gradlew` commands here to have `--no-daemon`.

carrotkite:
### JaCoCO Code Coverage 77.06% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/integration/FragmentFlowRenderView` | 64% | 0% | :white_check_mark: |


</p>

PR #144, opened 2020-06-11
Fix effect execution order.

Laimiux:
## What
Effects should be executed on a first come first serve basis. Currently, in some cases, it might execute them on last come first serve which can lead to unpredictable behavior. 

While adding a new test, I found out that by default, `assertThat().containsExactly()` doesn't assert on the order. To fix this we either need to append `.inOrder()` or call `isEqualTo(listOf())`. A few tests started failing once we started to assert on order.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
@@ -81,7 +81,7 @@ class FormulaRuntime<Input : Any, State, RenderModel : Any>(
                         threadChecker.check("Only thread that created it can trigger transitions.")
 
                         if (effects != null) {
-                            effectQueue.push(effects)
+                            effectQueue.addLast(effects)
                         }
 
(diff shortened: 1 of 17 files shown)


formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -81,7 +81,7 @@ class FormulaRuntime<Input : Any, State, RenderModel : Any>(
                         threadChecker.check("Only thread that created it can trigger transitions.")
 
                         if (effects != null) {
-                            effectQueue.push(effects)
+                            effectQueue.addLast(effects)
```
Laimiux, line 84: This is the fix: `push` appends the effect to the start of the list, we want to append to the end.

carrotkite:
### JaCoCO Code Coverage 77.07% :white_check_mark:
|:---|:---:|:---:|:---:|


</p>

PR #146, opened 2020-06-29
[docs] explain input equality.

Laimiux:
## What
Input equality is a common issue that comes up in PR reviews.

### DIFF
--- a/docs/input.md
+++ b/docs/input.md
@@ -47,3 +47,131 @@ itemDetailFormula
     
   }
 ```
+
+## Equality
+Formula uses input equality to determine if it should re-evaluate. A parent can cause 
+a child formula to re-evaluate by changing the input it passes. This will also trigger
+`Formula.onInputChanged` callback on the child formula.
+
+This is a desired behavior as we do want the child to react when the data that we pass changes.
+```kotlin
+data class ItemInput(
+    val itemId: String
+)
+```
+
+Making `Input` a data class and passing data as part of its properties makes it easy to reason
+about its equality. In some cases though, we want to pass objects that don't have property based
+equality such as callback functions or observables. 
+
+### Maintaining callback equality
+In many cases we want to pass callbacks to listen to formula events. 
+
+```kotlin
+data class ItemListInput(
+    val onItemSelected: (Item) -> Unit
+)
+```
+
+Formula provides an easy way to maintain callback equality. Within your parent formula, 
+you can use `FormulaContext` to instantiate callbacks by using:
+
+- `FormulaContext.callback`
+- `FormulaContext.eventCallback` 
+
+**Don't:** Don't instantiate functions within `Formula.evaluate`.
+```kotlin
+val itemListInput = ItemListInput(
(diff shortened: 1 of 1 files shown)


carrotkite:
### JaCoCO Code Coverage 76.85% :white_check_mark:
|:---|:---:|:---:|:---:|


</p>

docs/input.md
```diff
@@ -47,3 +47,131 @@ itemDetailFormula
...
+  val eventObservable: Observable<Event>
+)
+```
+
+**Don't:** create a new observable within `Formula.evaluate`
+```kotlin
+val input = MyInput(
+    eventObservable = relay.map { Event() }
+)
+```
```
Jawnnypoo, line 134: I don't think this is the correct example - this isn't creating a new observable within `evaluate`

Jawnnypoo: Should be:
```
override fun evaluate(...) {
    val input = MyInput(
        eventObservable = eventObservable
    )
}
```

docs/input.md
```diff
@@ -47,3 +47,131 @@ itemDetailFormula
...
+override fun initialState(input: Input) = State(
+  eventObservable = input.eventObservable.map { Event() }
+)
+
+override fun evaluate(...) {
+    val input = MyInput(
+        eventObservable = state.eventObservable
+    )
+}
+```
```
Jawnnypoo, line 163: Do we still want to use `State` if we are directly taking the Observable from `Input`, not modifying it and using its raw result within `events`? ie. 
```
   data class Input(
        val complementaryMatchesEventRelay: Observable<Boolean>
   )

    override fun evaluate(
        input: Input,
        state: State,
        context: FormulaContext<State>
    ): Evaluation<ICNotificationTrayRenderModel> {
        return Evaluation(
            renderModel = renderModel,
            updates = context.updates {
                events(input.complementaryMatchesEventRelay) {
                    transition(state.copy(complementaryMatches = it))
                }
            }
        )
    }
```

Laimiux: Using `Observable` directly from the `Input` is fine, the parent has to ensure equality.

Jawnnypoo:
A scenario that I run into quite frequently is that in order to create a `Stream` via `RxStream.fromObservable`, I need to use the `Input`. In this case, is it best to create an initial Stream with `Observable.empty()`, and then update it to the correct `Stream` in `onInputChange`, or is there some other way to accomplish this?

Laimiux:
> A scenario that I run into quite frequently is that in order to create a `Stream` via `RxStream.fromObservable`, I need to use the `Input`. In this case, is it best to create an initial Stream with `Observable.empty()`, and then update it to the correct `Stream` in `onInputChange`, or is there some other way to accomplish this?

Could you show an example of how you use the `Input`?

If you rely on data coming from the `Input`, you can use `cacheKey` to configure the right invalidation policy, for example 
```kotlin
RxStream.fromObservable(input.itemId) { repo.fetchitem(input.itemId) }
```

PR #147, opened 2020-06-30
Move start from extension to interface method

Jawnnypoo:
Presently, it is difficult to mock the entire output of a Formula using Mockito, due to `start` being an extension method, instead of an actual method on the `Formula` interface. Something like this:
```
val formula = mock<TaskListFormula>()
whenever(formula.start(any<TaskListFormula.Input>())).thenReturn(
   Observable.just(TaskListRenderModel(items = emptyList(), filterOptions = emptyList()))
)
```
would throw this exception:
```
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method com.instacart.formula.RuntimeExtensionsKt.start, parameter input
	at com.instacart.formula.RuntimeExtensionsKt.start(RuntimeExtensions.kt)
	at com.examples.todoapp.tasks.TaskListFormulaTest.test start override(TaskListFormulaTest.kt:40)
```
With the introduction of Java 8 compilation, we can move the methods within `RuntimeExtensions.kt` into the `Formula` class, and therefore mock them easily (see [here](https://github.com/instacart/formula/compare/jawn/start-as-method?expand=1#diff-e0909e85cf9a0b62ed86651c37c81281R39))

### DIFF
--- a/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
@@ -15,7 +15,7 @@ import java.util.LinkedList
 /**
  * Takes a [Formula] and creates an Observable<RenderModel> from it.
  */
-class FormulaRuntime<Input : Any, State, RenderModel : Any>(
+class FormulaRuntime<Input, State, RenderModel>(
     private val threadChecker: ThreadChecker,
     private val formula: Formula<Input, State, RenderModel>,
     private val childManagerFactory: FormulaManagerFactory,
@@ -26,7 +26,7 @@ class FormulaRuntime<Input : Any, State, RenderModel : Any>(
         /**
          * RuntimeExtensions.kt [start] calls this method.
          */
-        fun <Input : Any, State,  RenderModel : Any> start(
+        fun <Input, State,  RenderModel> start(
             input: Observable<Input>,
             formula: Formula<Input, State, RenderModel>,
             childManagerFactory: FormulaManagerFactory = FormulaManagerFactoryImpl()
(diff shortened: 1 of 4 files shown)


formula/src/main/java/com/instacart/formula/RuntimeExtensions.kt
```diff
@@ -1,17 +1,22 @@
+@file:Suppress("DeprecatedCallableAddReplaceWith")
```
Jawnnypoo, line 1: Supressing `ReplaceWith`, since the extension method is becoming a method on the class, there is no replacement to make, just a deletion of an import (which unfortunately, `ReplaceWith` does not cover).

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -153,4 +153,8 @@ class FormulaRuntime<Input : Any, State, RenderModel : Any>(
             }
         }
     }
+
+    private fun <T> checkAnyNotNull(value: T?): T {
+        return checkNotNull(value as Any?) as T
+    }
```
Jawnnypoo, line 159: This check is needed, since we have to change `Input : Any` and `RenderModel : Any`, to make `FormulaRuntime` compatible with the types defined within `Formula`. Otherwise, we would have to change over to `Input : Any` and `RenderModel : Any` in all classes that extend `Formula`.

carrotkite:
### JaCoCO Code Coverage 76.64% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/FormulaRuntime` | 87% | 0% | :white_check_mark: |


</p>

Laimiux:
I'm not certain about this change, will need to think about it for a bit.

Laimiux:
I was hoping to keep the runtime which exposes RxJava Observable out of the Formula interface. This would allow in the future to provide alternative runtimes such as co-routines.

Jawnnypoo:
Very good points, closing in favor of https://github.com/instacart/formula/pull/148

PR #148, opened 2020-06-30
Mockk extension example

Jawnnypoo:
Follow up to https://github.com/instacart/formula/pull/147. After doing some research, I found that while mockito does not do well with Kotlin extensions, [Mockk](https://github.com/mockk/mockk) is able to handle them well. This adds an example of providing a mock for a formula, overriding its output via mocking the extension. 

I think adding this as an example will be a good reference for those who might run into this.

### DIFF
--- a/samples/todoapp/src/test/java/com/examples/todoapp/tasks/TaskListFormulaTest.kt
+++ b/samples/todoapp/src/test/java/com/examples/todoapp/tasks/TaskListFormulaTest.kt
@@ -3,21 +3,21 @@ package com.examples.todoapp.tasks
 import com.examples.todoapp.data.Task
 import com.examples.todoapp.data.TaskRepo
 import com.google.common.truth.Truth.assertThat
+import com.instacart.formula.start
 import com.instacart.formula.test.test
-import com.nhaarman.mockito_kotlin.mock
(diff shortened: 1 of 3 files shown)


samples/todoapp/src/test/java/com/examples/todoapp/tasks/TaskListFormulaTest.kt
```diff
@@ -3,22 +3,24 @@ package com.examples.todoapp.tasks
...
 import io.reactivex.rxjava3.core.Observable
 import org.junit.Test
 
 class TaskListFormulaTest {
 
     @Test fun `change filter type`() {
-        val repo = mock<TaskRepo>()
-        whenever(repo.tasks()).thenReturn(Observable.just(
+        val repo = mockk<TaskRepo>()
+        every { repo.tasks() } returns Observable.just(
```
Laimiux, line 18: I would probably replace these with an interface fake instead of a mock.

carrotkite:
### JaCoCO Code Coverage 76.85% :white_check_mark:
|:---|:---:|:---:|:---:|


</p>

Jawnnypoo:
Closing in favor of https://github.com/instacart/formula/pull/149

PR #151, opened 2020-07-07
Potential fix for Java 8 build failure

Jawnnypoo:
Using `Disposible.fromRunnable` within Firebase Test Lab causes issues that are unknown at this time. See https://stackoverflow.com/questions/62762831/no-static-method-fromrunnable-when-running-tests-in-firebase-test-lab

We can likely undo this change later on, but I wanted to run tests with this in place and see if it fixes the issue for us.

### DIFF
--- a/formula/src/main/java/io/reactivex/rxjava3/disposables/FormulaDisposableHelper.kt
+++ b/formula/src/main/java/io/reactivex/rxjava3/disposables/FormulaDisposableHelper.kt
@@ -5,7 +5,7 @@ import java.util.Objects
 /**
  * Used due to issue with [Disposable.fromRunnable], please ignore outside of the internal library.
  */
-object FormulaDisposableHelper {
+internal object FormulaDisposableHelper {
 
     /**
      * Identical to [Disposable.fromRunnable]
--- a/formula/src/main/java/io/reactivex/rxjava3/disposables/FormulaDisposableHelper.kt
+++ b/formula/src/main/java/io/reactivex/rxjava3/disposables/FormulaDisposableHelper.kt
@@ -1,0 +1,17 @@
+package io.reactivex.rxjava3.disposables
+
+import java.util.Objects
+
+/**
+ * Used due to issue with [Disposable.fromRunnable], please ignore outside of the internal library.
+ */
+object FormulaDisposableHelper {
+
+    /**
+     * Identical to [Disposable.fromRunnable]
+     */
+    fun fromRunnable(run: Runnable): Disposable {
+        Objects.requireNonNull<Runnable>(run, "run is null")
+        return RunnableDisposable(run)
+    }
+}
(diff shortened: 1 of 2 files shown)


carrotkite:
### JaCoCO Code Coverage 76.82% :white_check_mark:
|:---|:---:|:---:|:---:|
| `io/reactivex/rxjava3/disposables/FormulaDisposableHelper` | 100% | 0% | :white_check_mark: |


</p>

formula/src/main/java/io/reactivex/rxjava3/disposables/FormulaDisposableHelper.kt
```diff
@@ -0,0 +1,17 @@
+package io.reactivex.rxjava3.disposables
+
+import java.util.Objects
+
+/**
+ * Used due to issue with [Disposable.fromRunnable], please ignore outside of the internal library.
+ */
+object FormulaDisposableHelper {
```
Laimiux, line 8: internal?

Jawnnypoo: If only. We can't, since we are pretending we are a part of the RxJava 3 package.

Laimiux: Do you need to pretend, maybe move it into our package?

Jawnnypoo: Yep, we have to pretend, because `RunnableDisposable` is package private.

Laimiux: I'm curious if it's enough to replace 
```kotlin
val disposable = Disposable.fromRunnable {
} 
```
with 
```kotlin
val runnable = Runnable {} 
val disposable = Disposable.fromRunnable(runnable)
```

Jawnnypoo: Good question. I don't think so, because `Disposible.fromRunnable` is a static interface method, which was added to Java 8 and therefore is likely the cause for this incompatibility with Firebase Test Lab

formula/src/main/java/io/reactivex/rxjava3/disposables/FormulaDisposableHelper.kt
```diff
@@ -5,7 +5,7 @@ import java.util.Objects
 /**
  * Used due to issue with [Disposable.fromRunnable], please ignore outside of the internal library.
  */
-object FormulaDisposableHelper {
+internal object FormulaDisposableHelper {
```
Jawnnypoo, line 8: Oh weird... I thought this would fail since it is a different package from where it is getting called...

PR #164, opened 2020-11-12
Emit error during evaluation or execution phase.

Laimiux:
## What
Let's catch errors and properly emit them as part of Observable callbacks.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
@@ -17,7 +17,8 @@ import java.util.LinkedList
 class FormulaRuntime<Input : Any, Output : Any>(
     private val threadChecker: ThreadChecker,
     formula: IFormula<Input, Output>,
-    private val onOutput: (Output) -> Unit
+    private val onOutput: (Output) -> Unit,
+    private val onError: (Throwable) -> Unit
 ) {
 
     companion object {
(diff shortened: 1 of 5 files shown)


carrotkite:
### JaCoCO Code Coverage 76.19% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/FormulaRuntime` | 84% | 0% | :white_check_mark: |


</p>

Jawnnypoo (approved):
Would this be considered a breaking change?

Laimiux:
> Would this be considered a breaking change?

Good question, I would say it's not exactly a breaking change - it's more of an improvement. For example, previously by calling `renderModel.onToolbarSelected()` you could unknowingly force a crash in the UI layer. We are now not crashing the UI thread, but passing it to the formula lifecycle owner through the observable callback.

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -116,21 +117,27 @@ class FormulaRuntime<Input : Any, Output : Any>(
...
 
-        if (hasInitialFinished && emitOutput) {
-            emitOutput = false
-            onOutput(checkNotNull(lastOutput))
+            if (hasInitialFinished && emitOutput) {
+                emitOutput = false
+                onOutput(checkNotNull(lastOutput))
+            }
+        } catch (e: Throwable) {
+            manager?.markAsTerminated()
```
Laimiux, line 138: I've added termination of formula when there is an error.

PR #178, opened 2020-12-23
Allow nullable formula states.

Laimiux:
## What
There is no reason why we shouldn't allow `null` states given that Kotlin type system provides support for it. This will remove the need to wrap nullable states within `Option` type.

- As part of this, I've separated `Transition` into 3 sealed class types. 
- Separated child transition logic - no need to initialize a new `Transition` object when passing it from child to parent.
- Only initialize child transition listener once.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
+++ b/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
@@ -45,9 +45,8 @@ internal class FormulaManagerImpl<Input, State, Output>(
             return
         }
 
-        this.state = when (transition) {
-            is Transition.Stateful -> transition.state
-            else -> this.state
+        if (transition is Transition.Stateful) {
+            state = transition.state
         }
         val frame = this.frame
         frame?.updateStateValidity(state)
@@ -198,16 +197,16 @@ internal class FormulaManagerImpl<Input, State, Output>(
 
     private fun getOrInitChildTransitionListener(): TransitionListener {
         return childTransitionListener ?: run {
-            val listener = TransitionListener { transition, isValid ->
+            TransitionListener { transition, isChildValid ->
                 val frame = this.frame
-                if (!isValid) {
+                if (!isChildValid) {
                     frame?.childInvalidated()
                 }
                 val isValid = frame != null && frame.isValid()
                 transitionListener.onTransition(transition, isValid)
+            }.apply {
+                childTransitionListener = this
             }
-            childTransitionListener = listener
-            listener
         }
     }
 }
--- a/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
+++ b/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
@@ -14,7 +14,7 @@ import com.instacart.formula.Transition
  * 3. Terminate removed children
  * 4. Prepare parent and alive children for updates.
  */
-internal class FormulaManagerImpl<Input, State : Any, Output>(
+internal class FormulaManagerImpl<Input, State, Output>(
     private val formula: Formula<Input, State, Output>,
     initialInput: Input,
     private val callbacks: ScopedCallbacks,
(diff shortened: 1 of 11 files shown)


formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -162,9 +163,7 @@ internal class FormulaManagerImpl<Input, State : Any, Output>(
         val compositeKey = constructKey(formula, input)
         val manager = children
             .findOrInit(compositeKey) {
-                val childTransitionListener = TransitionListener { effects, isValid ->
-                    handleTransition(Transition(effects = effects), !isValid)
```
Laimiux, line 166: Previously we initialized a new `Transition` when passing side-effects up. Depending on the number of layers, there could be a decent amount of unnecessary allocations.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -196,4 +195,19 @@ internal class FormulaManagerImpl<Input, State : Any, Output>(
             key = formula.key(input)
         )
     }
+
+    private fun getOrInitChildTransitionListener(): TransitionListener {
```
Laimiux, line 199: Optimization to only initialize `TransitionListener` once and re-use it for all child formulas.

carrotkite:
### JaCoCO Code Coverage 76.43% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/internal/TransitionUtils` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/FormulaManagerImpl` | 86% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/TransitionCallbackWrapper` | 75% | 0% | :white_check_mark: |
| `com/instacart/formula/FormulaRuntime` | 80% | 0% | :white_check_mark: |


</p>

JasonCromer (approved):
This LGTM so far. I want to spend a bit more time just digging into this tomorrow, but the change seems straightforward enough. I'm not sure I've spent enough time with Formula to foresee any unintentional side-effects at the moment. 
👍

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -36,22 +36,23 @@ internal class FormulaManagerImpl<Input, State : Any, Output>(
...
+    private fun handleTransition(transition: Transition<State>) {
         if (terminated) {
             // State transitions are ignored, only side effects are passed up to be executed.
-            transitionListener.onTransition(transition.effects, true)
+            transitionListener.onTransition(transition, true)
             return
         }
 
-        this.state = transition.state ?: this.state
+        this.state = when (transition) {
```
JasonCromer, line 48: Not sure how much we prefer optimizations vs readability, but you can always do the following to remove a few extra lines of bytecode to load/get the `this.state` field.
```
        if (transition is Transition.Stateful) {
            this.state = transition.state
        }
```

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -196,4 +195,19 @@ internal class FormulaManagerImpl<Input, State : Any, Output>(
...
     }
+
+    private fun getOrInitChildTransitionListener(): TransitionListener {
+        return childTransitionListener ?: run {
+            val listener = TransitionListener { transition, isValid ->
+                val frame = this.frame
+                if (!isValid) {
+                    frame?.childInvalidated()
+                }
+                val isValid = frame != null && frame.isValid()
```
JasonCromer, line 206: Looks like this val is shadowing the parent `isValid`. Not a big deal, but might be nice to change.

Laimiux: I thought I had fixed it... Renaming passed `isValid` to `isChildValid`

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -196,4 +195,19 @@ internal class FormulaManagerImpl<Input, State : Any, Output>(
             key = formula.key(input)
         )
     }
+
+    private fun getOrInitChildTransitionListener(): TransitionListener {
+        return childTransitionListener ?: run {
```
JasonCromer, line 200: You can return the `TransitionListener` object directly, probably more Kotlin idiomatic:
```
        return childTransitionListener ?: run {
            TransitionListener { transition, isValid ->
                val frame = this.frame
                if (!isValid) {
                    frame?.childInvalidated()
                }
                val isValid = frame != null && frame.isValid()
                transitionListener.onTransition(transition, isValid)
            }.apply {
                childTransitionListener = this
            }
        }
```

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
Laimiux: This looks better, updating...

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
Laimiux: I was debating if I should use this syntax.

JasonCromer (approved):

JasonCromer (approved):

PR #179, opened 2021-02-11
Fix potential unresponsive fragment issue when used with `FragmentContainerView`

mhartvick:
When using `FragmentContainerView` we might run in a situation where `onDestroyView` is called at the end of the animation.
If within that animation time frame we re-open the same fragment contract, it might get removed by the `onDestroyView()` call, and will not receive any state update anymore.

This addresses the problem by eagerly removing the fragment when it's being popped from the backstack right away (instead of waiting for the `onFragmentViewDestroyed()` and `onFragmentDetached()` callbacks)

### DIFF
--- a/formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
+++ b/formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
@@ -2,6 +2,7 @@ package com.instacart.formula.integration
 
 import android.content.Context
 import android.os.Bundle
+import android.util.Log
 import android.view.View
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentActivity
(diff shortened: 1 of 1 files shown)


carrotkite:
### JaCoCO Code Coverage 76.44% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/integration/FragmentFlowRenderView` | 64% | 0% | :white_check_mark: |


</p>

formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
```diff
@@ -2,6 +2,7 @@ package com.instacart.formula.integration
 
 import android.content.Context
 import android.os.Bundle
+import android.util.Log
```
Laimiux, line 5: nit: can we remove this import?

Laimiux (approved):
There was no way to add a robolectric test?

mhartvick:
> There was no way to add a robolectric test?

Nope, spent some time trying to reproduce the bug using the same scenario on a Roboelectric test but I wasn't able to do so.

PR #181, opened 2021-04-08
Fragment Feature API.

Laimiux:
## What
This is a change motivated by two parts:
- Allow dependency injection in the UI layer (aka moving view creation out of FragmentContract)
- Prepare for Compose UI toolkit (need to update FragmentContract API to support it)

What are we changing:
- Add support for  `FragmentKey` usage which replaces `FragmentContract` (Note: fragment contract will still be supported for quite some time)
```kotlin
@Parcelize
data class TaskListKey(
    override val tag: String = "task list",
) : FragmentKey
```
- New `ViewFactory` concept which moves UI logic out of `FragmentContract`
```kotlin
// We can make this view factory injectable
class TaskListViewFactory : LayoutViewFactory<TaskListRenderModel>(R.layout.task_list) {

    override fun ViewInstance.create(): FeatureView<TaskListRenderModel> {
        return featureView(TaskListRenderView(view))
    }
}
```

- New `Feature` and `FeatureFactory` which replace `Integration` 
```kotlin
class TaskListFeatureFactory : FeatureFactory<TaskListFeatureFactory.Dependencies, TaskListKey> {
    // Provided by the parent, which allows a form of dependency injection
    interface Dependencies {
        fun taskRepo(): TaskRepo
        fun taskListInput(): TaskListFormula.Input
    }

    override fun initialize(dependencies: Dependencies, key: TaskListKey): Feature<*> {
        // Note: we could create our own internal dagger component here using the dependencies.
        val formula = TaskListFormula(dependencies.taskRepo())
        return Feature(
            state = formula.toObservable(dependencies.taskListInput()),
            viewFactory = TaskListViewFactory()
        )
    }
}
```


Laimiux:
As a follow-up to this PR, I will update our documentation site with updated best practices (for example `Integration.md` file).

formula-android/src/main/java/com/instacart/formula/android/FeatureView.kt
```diff
@@ -0,0 +1,30 @@
...
+ * rendering. Formula fragment uses [ViewFactory.create] to instantiate [FeatureView].
+ *
+ * Use [ViewFactory.fromLayout] and [LayoutViewFactory] to define a [ViewFactory] which can create
+ * [FeatureView].
+ *
+ * @param view The root Android view.
+ * @param bind A bind function connects state observable to the view rendering.
+ * @param lifecycleCallbacks Optional lifecycle callbacks if you need to know the Fragment state.
+ */
+class FeatureView<RenderModel>(
```
Jawnnypoo, line 21: Do we still want to use the term `View` in so many places? Maybe `Screen` instead, `ScreenFactory`, `FeatureScreen`, etc? I am just wondering if we want to move away from `View` since Compose doesn't use Android Views, and though they are unrelated, they might be associated. Maybe even `FeatureUI`, `UIFactory`, etc? Naming is hard, so I understand if you disagree.

formula-android/src/main/java/com/instacart/formula/fragment/FragmentFlowStore.kt
```diff
@@ -87,12 +87,14 @@ class FragmentFlowStore(
...
                                 activeKeys = state.activeKeys.minus(key),
-                                states = state.states.minus(key)
+                                states = state.states.minus(key),
+                                features = state.features.minus(key)
                             )
                             transition(updated)
                         }
                         is LifecycleEvent.Added -> {
                             if (!state.activeKeys.contains(key)) {
+                                // TODO: should use input.onScreenError() instead of adding fake render model!
```
Jawnnypoo, line 97: Still todo?

samples/todoapp/src/main/java/com/examples/todoapp/TodoApp.kt
```diff
@@ -19,19 +17,11 @@ class TodoApp : Application() {
...
-                            bind(TaskListContract::class) { component, key ->
-                                val input = TaskListFormula.Input(showToast = { message ->
-                                    send {
-                                        onEffect(TodoActivityEffect.ShowToast(message))
-                                    }
-                                })
-
-                                component.createTaskListFormula().toObservable(input)
-                            }
+                            bind(TaskListFeatureFactory())
```
Jawnnypoo, line 24: So clean 👍

Jawnnypoo (approved):
This is a very welcome change to me 👍

formula-android/src/main/java/com/instacart/formula/fragment/FragmentFlowStore.kt
Laimiux: Yeah, I noticed this improvement while working on this feature, but will actually implement it in another PR.

formula-android/src/main/java/com/instacart/formula/android/ViewFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    /**
+     * This method is called from [FormulaFragment.onCreateView] function. Use it to
+     * instantiate an Android view instance and return a [FeatureView] which knows how to
+     * bind the state management to view rendering. Usually, you should use [LayoutViewFactory]
+     * or [ViewFactory.fromLayout] instead of implementing this method directly.
+     */
+    fun create(inflater: LayoutInflater, container: ViewGroup?): FeatureView<RenderModel>
```
alexanderbezverhni, line 55: Any reason making `container` nullable here?

formula-android/src/main/java/com/instacart/formula/android/ViewInstance.kt
```diff
@@ -0,0 +1,59 @@
...
+     * as part of [FormulaFragment.onCreateView].
+     */
+    abstract val view: View
+
+    /**
+     * Creates a [FeatureView] from a [render] function
+     */
+    fun <RenderModel> featureView(
+        lifecycleCallbacks: FragmentLifecycleCallback? = null,
+        render: (RenderModel) -> Unit
```
alexanderbezverhni, line 26: Worth swapping params order here to align with other `featureView()` counterparts.

formula-android/src/main/java/com/instacart/formula/fragment/FormulaFragment.kt
```diff
@@ -5,57 +5,53 @@ import android.view.LayoutInflater
...
 import androidx.fragment.app.Fragment
-import com.instacart.formula.RenderView
+import com.instacart.formula.Cancelable
+import com.instacart.formula.android.FeatureView
+import com.instacart.formula.android.ViewFactory
 import com.jakewharton.rxrelay3.BehaviorRelay
-import io.reactivex.rxjava3.disposables.Disposable
 
-class FormulaFragment<RenderModel> : Fragment(), BaseFormulaFragment<RenderModel> {
+class FormulaFragment : Fragment(), BaseFormulaFragment<Any> {
```
alexanderbezverhni, line 13: Why not to keep the narrowed `RenderModel` generic type?

alexanderbezverhni (approved):
Looks good! 💯

formula-android/src/main/java/com/instacart/formula/android/ViewFactory.kt
Laimiux: This is to align with `Fragment.onCreateView` which provides only nullable `container`. I also didn't like this value as nullable and looked at Fragment documentation to understand why and when it is nullable, however, there wasn't much information.

formula-android/src/main/java/com/instacart/formula/android/ViewInstance.kt
Laimiux: Having lambda as last param here allows better Kotlin structure such as 
```kotlin
featureView(lifecycleCallbacks) { renderModel ->
  // render
}
```

Laimiux: Might be worth updating the others to ask for lifecycle callbacks first, but it would make the other API usage not ideal. For example, if I reverse it to 
```kotlin
featureView(
  lifecycleCallbacks: FragmentLifecycleCallbacks? = null,
  renderView: RenderView
)
```
I likely won't be able to call it `featureView(MyRenderView())`

formula-android/src/main/java/com/instacart/formula/fragment/FormulaFragment.kt
Laimiux: We don't have any information about the generic value since we are receiving a `FragmentKey` which doesn't contain `RenderModel` type info. This actually brings a good point that it might make sense to clean up `FormulaFragment` and `BaseFormulaFragment` API surface further.

PR #190, opened 2021-04-23
[Fragments] Fixing a bug caused by multiple fragments with same key.

Laimiux:
## What
The bug is caused by adding multiple fragments with the same `FragmentKey` and `tag` to the back stack. The fragment integration makes an assumption that each fragment has a unique tag, but that is not necessarily true. To fix this, we need to start identifying fragments by something more than just their `FragmentKey`.


formula-android/src/main/java/com/instacart/formula/integration/ActiveFragment.kt
```diff
@@ -0,0 +1,8 @@
+package com.instacart.formula.integration
+
+import com.instacart.formula.fragment.FragmentKey
+
+data class ActiveFragment(
```
Laimiux, line 5: Need to figure out a better name for this.

formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
```diff
@@ -130,6 +133,21 @@ internal class FragmentFlowRenderView(
         }
     }
 
+    private fun initializeFormulaIdIfNeeded(f: Fragment) {
```
Laimiux, line 136: When a fragment is attached, we check if it has a generated id and generate one if it doesn't.

Laimiux: TODO: maybe rename to `initializeFragmentIdIfNeeded`

formula-android/src/main/java/com/instacart/formula/integration/LifecycleEvent.kt
```diff
@@ -1,10 +0,0 @@
-package com.instacart.formula.integration
-
-/**
- * Models when a key is attached and detached. Provides a way to indicate
- * when to initialize state stream and when to destroy it.
- */
-sealed class LifecycleEvent<out Key>(val key: Key) {
```
Laimiux, line 7: Deleting this generic type and creating a specific `FragmentLifecycleEvent`

formula-android/src/main/java/com/instacart/formula/fragment/FragmentFlowState.kt
```diff
@@ -11,10 +12,10 @@ import com.instacart.formula.integration.KeyState
  * @param states Last emitted state of each active [FragmentContract].
  */
 data class FragmentFlowState(
-    val activeKeys: List<FragmentKey> = emptyList(),
-    val visibleKeys: List<FragmentKey> = emptyList(),
-    val features: Map<FragmentKey, FeatureEvent> = emptyMap(),
-    val states: Map<FragmentKey, KeyState> = emptyMap()
+    val activeKeys: List<ActiveFragment> = emptyList(),
```
Laimiux, line 15: The riskiest part of this change is changing from `FragmentKey` type to `ActiveFragment` type (need to find a better name for this).

carrotkite:
### JaCoCO Code Coverage 77.16% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/fragment/FragmentFlowState` | 0% | 0% | :white_check_mark: |
| `com/instacart/formula/fragment/FragmentLifecycleEvent` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/fragment/FragmentFlowStore` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/fragment/FragmentLifecycle` | 70% | 0% | :white_check_mark: |
| `com/instacart/formula/fragment/FormulaFragment` | 50% | 0% | :white_check_mark: |
| `com/instacart/formula/integration/internal/FeatureBinding` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/integration/internal/CompositeBinding` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/android/internal/ActivityStoreContextImpl` | 80% | 0% | :white_check_mark: |
| `com/instacart/formula/integration/FragmentFlowRenderView` | 77% | 0% | :white_check_mark: |
| `com/instacart/formula/integration/FormulaFragmentViewFactory` | 60% | 0% | :white_check_mark: |
| `com/instacart/formula/integration/ActiveFragment` | 66% | 0% | :white_check_mark: |
| `com/instacart/formula/integration/Binding` | 100% | 0% | :white_check_mark: |


</p>

Laimiux:
@mhartvick I think this change might allow us to get rid of the early removal from the backstack.

formula-android-tests/src/main/java/com/instacart/formula/test/TestFragmentActivity.kt
```diff
@@ -19,7 +21,36 @@ class TestFragmentActivity : FormulaAppCompatActivity() {
             val fragment = FormulaFragment.newInstance(initialContract)
             supportFragmentManager.beginTransaction()
                 .add(R.id.activity_content, fragment, initialContract.tag)
+                .addToBackStack(initialContract.tag)
                 .commit()
         }
     }
+
+    fun navigateTo(key: FragmentKey) {
```
Laimiux, line 29: Improving the API for testing fragment navigation.

formula-android-tests/src/test/java/com/instacart/formula/FragmentFlowRenderViewTest.kt
```diff
@@ -184,18 +184,37 @@ class FragmentFlowRenderViewTest {
         assertThat(activeContracts()).containsExactly(TestContract()).inOrder()
     }
 
+    @Test fun `add multiple fragments with same fragment key`() {
```
Laimiux, line 187: This is the test that was used to replicate production scenario.

formula-android-tests/src/test/java/com/instacart/formula/FragmentFlowRenderViewTest.kt
```diff
@@ -184,18 +184,37 @@ class FragmentFlowRenderViewTest {
...
 
-    private fun navigateToTaskDetail() {
-        val detail = TestContractWithId(1)
+    private fun navigateToTaskDetail(id: Int = 1) {
+        scenario.onActivity {
+            it.navigateTo(TestContractWithId(id))
+        }
+    }
+
+    private fun assertFragmentViewIsCreated(key: FragmentKey) {
```
Laimiux, line 214: Adding a check to make sure that view is actually created (making sure that `getFeature` is called).

formula-android/src/main/java/com/instacart/formula/android/internal/ActivityStoreContextImpl.kt
```diff
@@ -104,7 +105,9 @@ internal class ActivityStoreContextImpl<Activity : FragmentActivity> : ActivityS
         attachEventRelay.accept(false)
     }
 
-    fun updateFragmentLifecycleState(contract: FragmentKey, newState: Lifecycle.State) {
+    fun updateFragmentLifecycleState(id: FragmentId, newState: Lifecycle.State) {
+        // TODO: should probably start using [id] instead of [contract] here.
```
Laimiux, line 109: This is a bigger rethink as in other places we are listening for fragment events using only `FragmentKey`. It's a much smaller issue than this, so will perform these fixes in a follow-up.

formula-android/src/main/java/com/instacart/formula/fragment/FragmentFlowState.kt
```diff
@@ -11,10 +12,18 @@ import com.instacart.formula.integration.KeyState
...
-    val features: Map<FragmentKey, FeatureEvent> = emptyMap(),
-    val states: Map<FragmentKey, KeyState> = emptyMap()
+    val activeIds: List<FragmentId> = emptyList(),
+    val visibleIds: List<FragmentId> = emptyList(),
+    val features: Map<FragmentId, FeatureEvent> = emptyMap(),
+    val states: Map<FragmentId, KeyState> = emptyMap()
 ) {
-    fun visibleState() = visibleKeys.lastOrNull()?.let { states[it] }
+    @Deprecated("use activeIds")
+    val activeKeys: List<FragmentKey>
```
Laimiux, line 21: Instead of breaking the API, I'm leaving `activeKeys` and `visibleKeys` for now.

mhartvick:
> @mhartvick I think this change might allow us to get rid of the early removal from the backstack logic.

💯  we should be able to remove it now. I'll need to double check this was not fixing any other side effect (but wouldn't expect so)

formula-android-tests/src/test/java/com/instacart/formula/FragmentFlowRenderViewTest.kt
```diff
@@ -184,18 +184,37 @@ class FragmentFlowRenderViewTest {
...
+        navigateToTaskDetail(id = 1)
+
+        assertFragmentViewIsCreated(TestContractWithId(1))
+        assertThat(activeContracts()).containsExactly(TestContract(), TestContractWithId(1), TestContractWithId(2), TestContractWithId(1)).inOrder()
+
+        navigateBack()
+        navigateBack()
+
+        assertFragmentViewIsCreated(TestContractWithId(1))
+        assertThat(activeContracts()).containsExactly(TestContract(), TestContractWithId(1)).inOrder()
```
alexanderbezverhni, line 201: 💯

formula-android/src/main/java/com/instacart/formula/fragment/FragmentFlowStore.kt
```diff
@@ -132,11 +132,11 @@ class FragmentFlowStore(
                     if (feature != null) {
                         RxStream.fromObservable(feature) {
                             feature.state.onErrorResumeNext {
-                                input.onScreenError(key, it)
+                                input.onScreenError(key.key, it)
                                 Observable.empty()
                             }
                         }.onEvent {
-                            val keyState = KeyState(key, it)
+                            val keyState = KeyState(key.key, it)
```
alexanderbezverhni, line 139: nit: `key` -> `framentId`, to avoid `key.key` ambiguity?

formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
```diff
@@ -198,9 +198,28 @@ internal class FragmentFlowRenderView(
...
+    private fun initializeFragmentInstanceIdIfNeeded(f: Fragment) {
+        if (f is BaseFormulaFragment<*>) {
+            val arguments = f.arguments ?: run {
+                Bundle().apply {
+                    f.arguments = this
+                }
+            }
+            val id = arguments.getString(FormulaFragment.ARG_FORMULA_ID, "")
+            if (id.isNullOrBlank()) {
+                val initializedId = UUID.randomUUID().toString()
```
alexanderbezverhni, line 220: Maybe lookup for existing fragments `ARG_FORMULA_ID`, to avoid collision?

formula-android/src/test/java/com/instacart/formula/fragment/FragmentFlowStoreTest.kt
```diff
@@ -55,15 +55,15 @@ class FragmentFlowStoreTest {
         val detail = Detail(1)
         store
             .state(FragmentEnvironment())
-            .map { it.states }
+            .map { it.states.mapKeys { it.key.key } }
```
alexanderbezverhni, line 58: `key.key` reads weirdly

alexanderbezverhni (approved):
Few nits, LGTM otherwise 👍🏼

formula-android/src/test/java/com/instacart/formula/fragment/FragmentFlowStoreTest.kt
Laimiux: Agreed, I've updated it to `entry.key.key` though it only makes it slightly better.

formula-android/src/main/java/com/instacart/formula/integration/FragmentFlowRenderView.kt
Laimiux: This is an interesting thought which I had to read up more on.

From [Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Random_UUID_probability_of_duplicates)
```
While the probability that a UUID will be duplicated is not zero, it is close enough to zero to be negligible
```

Also, from [Stackoverflow](https://stackoverflow.com/questions/24876188/how-big-is-the-chance-to-get-a-java-uuid-randomuuid-collision)

Screen Shot 2021-04-26 at 11 29 03 AM


Also, since we use this in-combination with `FragmentKey`, it's even less change for a duplicate.

alexanderbezverhni: This 1 crash in 100 years from now is going to be on you 🤣

PR #208, opened 2021-05-22
[Android] Initial logger.

Laimiux:
## What
Adding a hook to access debug `formula-android` events. I will provide more events in the future, but for now, I'm only interested in ignored attach events.

### DIFF
--- a/formula-android/src/main/java/com/instacart/formula/FormulaAndroid.kt
+++ b/formula-android/src/main/java/com/instacart/formula/FormulaAndroid.kt
@@ -26,7 +26,7 @@ object FormulaAndroid {
      */
     fun init(
         application: Application,
-        logger: (String) -> Unit = {},
+        logger: ((String) -> Unit)? = null,
         onFragmentError: (FragmentKey, Throwable) -> Unit = { _, it -> throw it },
         activities: ActivityConfigurator.() -> Unit
     ) {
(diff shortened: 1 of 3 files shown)


carrotkite:
### JaCoCo Code Coverage 76.93% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/android/FragmentEnvironment` | 66% | 0% | :white_check_mark: |
| `com/instacart/formula/android/internal/FragmentFlowRenderView` | 77% | 0% | :white_check_mark: |
| `com/instacart/formula/FormulaAndroid` | 50% | 0% | :white_check_mark: |


</p>

formula-android/src/main/java/com/instacart/formula/FormulaAndroid.kt
```diff
@@ -21,10 +21,12 @@ object FormulaAndroid {
     /**
      * Initializes Formula Android integration. Should be called within [Application.onCreate].
      *
+     * @param logger A logger for debug Formula Android events.
      * @param onFragmentError A global handler for fragment errors. Override this to log the crashes.
      */
     fun init(
         application: Application,
+        logger: (String) -> Unit = {},
```
Jawnnypoo, line 29: Should it be nullable if the user doesn't care to log? and if so, maybe it should be the last param?

Jawnnypoo (approved):

formula-android/src/main/java/com/instacart/formula/FormulaAndroid.kt
Laimiux: I've made the `logger` nullable, however, it cannot be the last parameter because the following continuation is already occupying that spot.
```
activities: ActivityConfigurator.() -> Unit
```

PR #212, opened 2021-06-09
Add :formula-android-compose module.

Laimiux:
## What
This module will provide utilities to integrate Formula, Android and Compose.


dependencies.gradle
```diff
@@ -20,6 +21,18 @@ ext {
                     core            : [
                             ktx: "androidx.core:core-ktx:1.2.0"
                     ],
+                    activity        : [
+                            compose : "androidx.activity:activity-compose:1.3.0-alpha05"
+                    ],
+                    compose         : [
+                            activity    : "androidx.activity:activity-compose:1.3.0-alpha05",
+                            foundation  : "androidx.compose.foundation:foundation:$composeVersion",
+                            material    : "androidx.compose.material:material:$composeVersion",
```
FrancoisBlavoet, line 30: why would you need material in formula ?

formula-android-compose/src/main/java/com/instacart/formula/android/compose/ComposeViewFactory.kt
```diff
@@ -0,0 +1,32 @@
...
+abstract class ComposeViewFactory<RenderModel> : ViewFactory<RenderModel> {
+
+    override fun create(inflater: LayoutInflater, container: ViewGroup?): FeatureView<RenderModel> {
+        val view = ComposeView(inflater.context)
+        return FeatureView(
+            view = view,
+            bind = {
+                view.setContent {
+                    val model = it.observable.subscribeAsState(null).value
+                    if (model != null) {
```
FrancoisBlavoet, line 21: I have been wondering if that's going to be an issue (like a binking first frame).
Hopefully not, having to provide a "default" UI would be inconvernient
Reworking Formula so the RenderModelGenerator emits synchronously the first time would probable be a large effort

FrancoisBlavoet (commented):
looks good, I think that's the right approach. Pretty much what I had in mind as well, I just wonder if we will have issues with that first frame 🤞

carrotkite:
### JaCoCo Code Coverage 77.75% :white_check_mark:
|:---|:---:|:---:|:---:|


</p>

dependencies.gradle
Laimiux: It's used in `samples/stopwatch-compose`

formula-android-compose/src/main/java/com/instacart/formula/android/compose/ComposeViewFactory.kt
Laimiux: Only legacy `RxFormula` might not have a default emission and we don't have too many of those. The `Formula` instances always emit and render model should already be available when `bind` is called. However, I'll confirm this behavior once I start testing the API.

formula-android-compose/src/main/java/com/instacart/formula/android/compose/ComposeViewFactory.kt
```diff
@@ -0,0 +1,32 @@
...
+                        Content(model)
+                    }
+                }
+                null
+            }
+        )
+    }
+
+    @Composable
+    abstract fun Content(model: RenderModel)
```
FrancoisBlavoet, line 31: one thing that would be nice to add (on our side, not in formula) will be to automatically call `DesignTheme` as the root composable, so that consumers of formula + IDS don't have to deal with this boilerplate

formula-android-compose/src/main/java/com/instacart/formula/android/compose/ComposeViewFactory.kt
FrancoisBlavoet: the proof will be in the pudding then 🤞

samples/stopwatch-compose/src/main/java/com/instacart/formula/compose/stopwatch/StopwatchActivity.kt
```diff
@@ -1,69 +1,19 @@
...
+class StopwatchActivity : FormulaAppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContent { render(StopwatchFormula().toObservable()) }
-    }
-}
+        setContentView(R.layout.stopwatch_activity)
 
-@Composable
-fun render(state: Observable<StopwatchRenderModel>) {
```
Laimiux, line 35: Moving all this logic into `StopwatchViewFactory`

formula-android-compose/src/main/java/com/instacart/formula/android/compose/ComposeViewFactory.kt
Laimiux: I tested this with fragments and it works fine to skip the null value.

FrancoisBlavoet (approved):
LGTM 💯 
Excited to start toying with this !

PR #220, opened 2021-08-06
updating index to fix the quickstart so that it actually works.

lgleasain:
I found a bunch of broken stuff in the index documents for Formula. If you try following the example listed in the quick start it also 
does not work and is different from the sample project it references. This PR updates that. I also created a separate project here https://github.com/lgleasain/counterexample that reflects what index guides you through creating. We will want to bring that into Formula or as a separate Instacart Public repo. Please let me know if I missed anything or if you have any ideas etc..

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,16 +1,31 @@
 Formula is a functional reactive framework built in Kotlin for managing state and side effects. It enables building 
 deterministic, composable, testable applications.
 
-### Quick example
-To demonstrate some of the concepts we will make a simple counter application. It shows the current count and 
+### Quick Start
+To demonstrate some of the concepts you will make a simple counter application. It shows the current count and 
 has two buttons to increment and decrement it.
 
 One of the best practises when working with Formula is to first think what the UI needs for rendering and what 
 actions the user will be able to take. This concept is called a `RenderModel` and is represented by a Kotlin data class.
 
+### Project Setup
+To get started create a new Android application called `Counter Application` with a blank activity. Next, rename (via refactor) 
+your `MainActivity` to `MyActivity`. Note: this is to make it easier for you to copy and past the example code later on.
+
+### Dependencies
+Now that you have your project set up, add the following libraries to your app level gradle file:
+
+```groovy
+dependencies {
+  implementation 'com.instacart.formula:formula-rxjava3:0.7.0'
+  implementation 'com.instacart.formula:formula-android:0.7.0'
+}
+```
+
 #### Render Model
 Render Model is an immutable representation of your view. It will be used to update Android views. Typically,
-it will also contain callbacks that will be invoked when user interacts with the UI.
+it will also contain callbacks that will be invoked when user interacts with the UI. Create a file called
+`CounterRenderModel` in the same package as your activity and add the following content.
 ```kotlin
 data class CounterRenderModel(
   val count: String,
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -1,16 +1,31 @@
...
-To demonstrate some of the concepts we will make a simple counter application. It shows the current count and 
+### Quick Start
+To demonstrate some of the concepts you will make a simple counter application. It shows the current count and 
 has two buttons to increment and decrement it.
 
 One of the best practises when working with Formula is to first think what the UI needs for rendering and what 
 actions the user will be able to take. This concept is called a `RenderModel` and is represented by a Kotlin data class.
 
+### Project Setup
+To get started create a new Android application called `Counter Application` with a blank activity. Next, rename (via refactor)
```
Laimiux, line 12: Maybe instead of this, I would recommend cloning the repository and going into `samples/counter`

docs/index.md
```diff
@@ -1,16 +1,31 @@
...
+dependencies {
+  implementation 'com.instacart.formula:formula-rxjava3:0.7.0'
+  implementation 'com.instacart.formula:formula-android:0.7.0'
+}
+```
+
 #### Render Model
 Render Model is an immutable representation of your view. It will be used to update Android views. Typically,
-it will also contain callbacks that will be invoked when user interacts with the UI.
+it will also contain callbacks that will be invoked when user interacts with the UI. Create a file called
```
Laimiux, line 27: I wonder if we really need to be so explicit with `Create a file called
CounterRenderModel in the same package as your activity and add the following content.`

docs/index.md
```diff
@@ -178,19 +238,41 @@ class MyApp : Application() {
...
 ```
 
+This over-rides `onCreate()` to call `init` on FormulaAndroid. The important thing here is that you have a stream that creates an new
+instance of your `CounterFormula` and calls update passing in the formula as an observable and a reference
+to the `render` function in your Activity.
+
+Next you will need to update your application in your `manifest.xml` to reference your new app class. Your new one 
+should look like this:
+```xml
+  <application
```
Laimiux, line 248: I would probably avoid introducing such details as most Android devs should know how to register `Application`.

docs/index.md
```diff
@@ -45,42 +63,88 @@ class CounterRenderView(root: ViewGroup): RenderView<CounterRenderModel> {
...
 
-We now defined a single entry-point to our rendering (this makes debugging issues a lot easier). Anytime 
-you need to update UI, just set a new Render Model.
+This is createing a `Renderer` that has a TextView for displaying the current count value state along with a
+increment and decrement button with click listeners that call the `onDecrement()` and `onIncrement()` methods
+you created in your `CounterRenderModel`. This codce also references some view elements that you don't 
+currently have. To fix that replace your activity view with the following.
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
```
Laimiux, line 72: I wonder if we should include this whole XML, makes the example pretty dense. Maybe it would be enough to include a screenshot of the UI and provide a link to XML layout.

lgleasain: If we want to have an example that a user can follow along with, unless we have a starter project, this is the only other way to do that given that these are referenced in our examples.

docs/index.md
lgleasain: I could change this to just mention to remember to update your application to use the new MyApp class.

Issue #231, opened 2021-09-10
Add Kotlin coroutine support.


Laimiux:

Laimiux:
Coroutine support has been added, there are still couple of things remaining here:
- we need to actually release `formula-coroutines` module to maven
- we need to document the APIs.

PR #233, opened 2021-09-11
Coroutine flow support

emmano:
This should help with adding support for coroutines as mentioned in #231 

A couple of things to note:

- I had to use the experimental `channel` field in `callbackFlow{}` in order to comply to the `onError` argument for `FormulaRuntime`. Using `cancel()` on the `ProducerScope` is not possible since it takes a `CancellableException`  and not a `Throwable`.
- The documentation for ViewModel integration suggests to dispose of the "producer" observable inside `onClear()` and to dispose of the consumer subscirption inside `onDestroy()` of the Activity. If I am not mistaken it seems that this can lead to two dangerous scenarios: 
- 1. The producer stream can trigger UI updates when the Activity/Fragment is not visible. This can be solved by disposing of the consumer inside `onStop()`.
- 2. The producer stream might be wasting resources since it is kept alive until the ViewModel is cleared. For example, if the producer stream depends on a stream that monitors network connectivity, the producer stream will still emit events even if the consumer has been disposed, hence wasting resources (i.e. battery). To address this issue, in one of the samples I added, I am using `repeatOnLifecycle{}` in combination with `SharingStarted.WhileSubscribed()`. This approach not only stops/starts collection, but also cancels the producer on the required lifecycle when there are no more collectors. `repeatOnLifecycle{}` uses an alpha release of the lifecycle runtime dependency.
- I added a test for `StopWatchFormula` just as an example of how testing a flow might look like. It is probably better to add Flow testing capabilities to `formula-test` just like it is done for rx.


formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowRuntime.kt
```diff
@@ -0,0 +1,38 @@
...
+
+
+            var runtime = FormulaRuntime(threadChecker, formula, ::trySendBlocking, channel::close)
+
+            input.collect { input ->
+                threadChecker.check("Input arrived on a wrong thread.")
+                if (!runtime.isKeyValid(input)) {
+                    runtime.terminate()
+                    runtime =
+                        FormulaRuntime(threadChecker, formula, ::trySendBlocking, channel::close)
```
Laimiux, line 27: Would `channel::close` trigger `awaitClose` callback which would call `runtime.terminate()` again?

emmano: It might. This should be better:

```kotlin
if(!channel.isClosedForSend) {
    runtime.terminate()
}
```

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowStream.kt
```diff
@@ -0,0 +1,82 @@
...
+         * ```
+         * events(FlowStream.fromFlow(itemId) { repo.fetchItem(itemId) }) { event ->
+         *   transition()
+         * }
+         * ```
+         *
+         * @param key Used to distinguish this [Stream] from other streams.
+         */
+        inline fun <Message> fromFlow(
+            scope: CoroutineScope,
```
Laimiux, line 56: Could we provide a default global scope here? The lifecycle of `Stream` will be controlled by `Formula` so in most cases, we don't really need to control it using scope.

samples/stopwatch-coroutines/src/main/java/com/instacart/formula/stopwatch/StopWatchViewModel.kt
```diff
@@ -0,0 +1,16 @@
+package com.instacart.formula.stopwatch
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.instacart.formula.coroutines.toFlow
+import kotlinx.coroutines.flow.*
```
Laimiux, line 6: nit: we generally don't use wildcard imports.

samples/stopwatch-coroutines/src/main/java/com/instacart/formula/stopwatch/StopWatchViewModel.kt
```diff
@@ -0,0 +1,16 @@
...
+
+class StopWatchViewModel(private val formula: StopwatchFormula = StopwatchFormula()) : ViewModel() {
+
+    val rendererStream by lazy {
```
Laimiux, line 10: Generally, we want to keep the formula running until the screen is destroyed. It seems with `SharingStarted.WhileSubscribed`, configuration changes would reset the state which is not what we usually want.

samples/stopwatch-coroutines/src/test/java/com/instacart/formula/stopwatch/StopwatchFormulaTest.kt
```diff
@@ -0,0 +1,33 @@
...
+@ExperimentalStdlibApi
+class StopwatchFormulaTest : CoroutineTest {
+
+    @get:Rule
+    override val coroutineRule = CoroutineTestRule()
+
+    @Test
+    fun `increment 5 times`() = test {
+
+        StopwatchFormula(this)
```
Laimiux, line 21: It would be good to have such a tests within `formula-coroutines` module to cover `toFlow()` and `FlowStream.fromFlow {}`.

samples/stopwatch-coroutines/src/main/java/com/instacart/formula/stopwatch/StopWatchViewModel.kt
emmano: Does this mean that a formula that reacts to a stream of events like the `NetworkStream` sample will keep the producer stream alive even if the Fragment is not visible?

samples/stopwatch-coroutines/src/main/java/com/instacart/formula/stopwatch/StopWatchViewModel.kt
emmano: I agree. I blame Android Studio for this. 😂

Laimiux:
First of all, thank you for this contribution 🙌 Overall, it looks pretty solid.

> * The documentation for ViewModel integration suggests to dispose of the "producer" observable inside `onClear()` and to dispose of the consumer subscirption inside `onDestroy()` of the Activity. 

We generally want `Formula` to survive configuration changes and forward navigation.
- We want to maintain state across configuration changes.
- We want network requests to persist across configuration changes as well.

This approach has been working well for us, but the issues you bring up are valid.
- For UI updates, we control the subscription from `Activity` or `Fragment`, so we can unsubscribe even though `Formula` is still running.
- Regarding wasteful/heavy resources, we haven't run into this issue much, but one option is to control it at the resource level. Formula keeps running, but the resource stream receives `Activity` lifecycle events and pauses when app is in the background.

> I added a test for StopWatchFormula just as an example of how testing a flow might look like. It is probably better to add Flow testing capabilities to formula-test just like it is done for rx.

I have been thinking about this a little bit, not yet sure what is the best option here. For more context, the RxJava version is used only internally within `formula-test`. My current thinking is that tests shouldn't really care about which runtime is used as they should behave equivalently. One caveat is that Coroutines has better support for nullable inputs and outputs where `RxJava` has to force non-null values currently.

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowRuntime.kt
```diff
@@ -0,0 +1,38 @@
+package com.instacart.formula.coroutines
+
+import com.instacart.formula.FormulaRuntime
+import com.instacart.formula.IFormula
+import com.instacart.formula.internal.ThreadChecker
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.channels.trySendBlocking
+import kotlinx.coroutines.flow.*
```
Laimiux, line 8: nit: explicit imports are preferable to wildcard imports.

samples/counter-coroutines/src/main/java/com/instacart/formula/counter/CounterActivity.kt
```diff
@@ -0,0 +1,32 @@
+package com.instacart.formula.counter
+
+import android.os.Bundle
+import androidx.fragment.app.FragmentActivity
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
+import com.instacart.formula.coroutines.toFlow
+import kotlinx.coroutines.flow.*
```
Laimiux, line 9: Nit: explicit imports preferable over wildcard.

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowStream.kt
emmano: I thought about this. Using coroutine's own `GlobalScope` is mainly considered bad practice. I am just not sure how to create a `Scope` for a formula since its lifecycle is really controlled by Android. For my example `StopwatchFormula` uses `MainScope` which uses a `SupervisorJob` + `Dispatchers.Main`. I could move this default into `fromFlow`.

samples/stopwatch-coroutines/src/main/java/com/instacart/formula/stopwatch/StopWatchViewModel.kt
Laimiux: Yes, it would keep it alive. You could add logic within `Formula` to optimize expensive operations to detach
```kotlin
if (state.isFragmentInFocus) {
    NetworkStream().onEvent {}
}
```

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowStream.kt
Laimiux: `MainScope` seems like a decent default.

samples/stopwatch-coroutines/src/test/java/com/instacart/formula/stopwatch/StopwatchFormulaTest.kt
emmano: I added test for `FlowRuntime` and `FlowStream`. I know you were trying to decide if it still made sense to have per runtime specific tests, but we can always remove them. I added `TestFlow` to serve the purpose of Rx's `TestObserver`. I know square came out with [Turbine](https://github.com/cashapp/turbine), but it felt like an overkill.

Laimiux:
Thank you for your contribution @emmano!! I'm going to take it over from here, finish up and merge the changes in #240.

emmano:
> Thank you for your contribution @emmano!! I'm going to take it over from here, finish up and merge the changes in #240.

No problem! Glad to help!

Laimiux:
Closing this as we merged #240

PR #236, opened 2021-09-15
[Docs] Adding a section about declarative actions and formula internals.

Laimiux:
## What
Adding documentation to cover the declarative nature of Formula and provide a simplified mental model of how actions are handled. This is not exhaustive yet, but hopefully a decent starting point.

### DIFF
--- a/docs/Declarative-API.md
+++ b/docs/Declarative-API.md
@@ -1,0 +1,80 @@
+Formula is a mix of functional, reactive and declarative programming. One aspect that might seem quite unusual is the
+way that it handles asynchronous actions such as RxJava observables, Kotlin Flows, etc. Most developers are used to
+explicitly managing subscription lifecycle.
+```kotlin
+val fetchUserObservable = repository.fetchUser()
+disposables += fetchUserObservable.subscribe { userResult ->
+    // Do something
+}
+```
+
+Formula does things a bit differently. It manages the lifecycle of the asynchronous actions for you. Instead of manually subscribing and unsubscribing,
+you define the conditions for which the asynchronous action should run and the callback which handles events produced by the action.
+```kotlin
+val fetchUserStream = RxStream.fromObservable { repository.fetchUser() }
+fetchUserStream.onEvent { userResult ->
+  // Do something
+}
+```
+
+The logic looks very similar to the first option, but the key difference here is that `Observable.subscribe` hasn’t run yet - the execution
+is deferred. It might not be clear why this is useful just from this example, but deferring execution allows us to provide a declarative API.
+For example, we can add conditional logic to only fetch user when user id is set.
+```kotlin
+if (state.userId != null) {
+    val fetchUserStream = RxStream.fromObservable { repository.fetchUser(state.userId) }
+    fetchUserStream.onEvent { userResult ->
+        // Do something with the result
+    }
+}
+```
+
+What if we want to fetch user information only after user clicks on some button to enable this. We can just expand on the conditional logic.
+```kotlin
+if (state.isUserFetchEnabled && state.userId != null) {
+    // Logic here is the same as the previous example
+}
+```
+
+Here, we don’t care what controls `isUserFetchEnabled` boolean. Formula will start execution when these
+conditions are met and will dispose of the action if `state.isUserFetchedEnabled` becomes `false` again.
+
+What if for some unusual reason the userId could change and we would want to refetch? We can
+define this behavior using `key` parameter
+```kotlin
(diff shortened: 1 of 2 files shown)


carrotkite:
### JaCoCo Code Coverage 77.09% :white_check_mark:
|:---|:---:|:---:|:---:|


</p>

docs/Declarative-API.md
```diff
@@ -0,0 +1,80 @@
...
+        // Do something with the result
+    }
+}
+```
+
+To understand how all this works, we will make some simplified assumptions about Formula APIs:
+
+- Each `Formula` will define an immutable `State` data class.
+- We use `State` object in the `evaluate(state: State)` function which defines the `Pair<UIModel, Actions>`
+- Any time there is a `State` change, the Formula runtime will call `evaluate` again.
```
jhowens89, line 57: `evaluate` is called on an input change too, right? I don't think you should dive into that completely here, but a quick "Any time there is a `State` change or `Input` change" could help reinforce the mental model.

docs/Declarative-API.md
```diff
@@ -0,0 +1,80 @@
...
+```kotlin
+if (state.isUserFetchEnabled && state.userId != null) {
+    val fetchUserStream = RxStream.fromObservable(key = state.userId) { repository.fetchUser(state.userId) }
+    fetchUserStream.onEvent { userResult ->
+        // Do something with the result
+    }
+}
+```
+
+To understand how all this works, we will make some simplified assumptions about Formula APIs:
```
jasonostrander, line 53: I think there should be more info here that this is a sample internal Formula implementation, and not application code. Maybe create a new heading in this file (like "Formula internals" or something) and add a sentence to make it clear.

jasonostrander (commented):
Looks great! I would also suggest to add some more on how the output is declarative and based on state+input. So far, this just covers how formula deals with async updates.

docs/Declarative-API.md
lgleasain: I'm wondering what happens here. Does this stop observing events for the previous key?

Laimiux: Yeah, Formula runtime disposes of old observables.

docs/Declarative-API.md
```diff
@@ -0,0 +1,104 @@
...
+    // Start actions
+    actions.forEach { action ->
+        if (!actionCache.contains(key)) {
+            actionCache[key] = action.start()
+        }
+    }  
+}
+```
+
+It's worth mentioning that this is an approximate and not the actual implementation. Within our
```
Laimiux, line 82: @jhowens89 @jasonostrander I've expanded below on both of your suggestions.

jhowens89 (approved):

jasonostrander (approved):
Looks good to me!

docs/Declarative-API.md
```diff
@@ -0,0 +1,104 @@
...
+```kotlin
+if (state.isUserFetchEnabled && state.userId != null) {
+    // Logic here is the same as the previous example
+}
+```
+
+Here, we don’t care what controls `isUserFetchEnabled` boolean. Formula will start execution when these
+conditions are met and will dispose of the action if `state.isUserFetchedEnabled` becomes `false` again.
+
+What if for some unusual reason the userId could change and we would want to refetch? We can
```
rjaylwar, line 42: We might want to change this a little to clarify that by default even though the evaluate function can be run may times, our stream will stay the same and only be subscribed when it is first added to the evaluate output. When I first encountered formula code that did something like RxStream.fromObservable { Observable.timer(...) }, I wasn't immediately sure that formula would make sure my timer wouldn't get restarted each time the evaluate function ran.

Maybe something along these lines to make everything explicit...  
By default because the `evaluate` function can be called many times, the `Stream` is only subscribed to when it is first added to the `evaluate` output. We can modify this behavior by using the `key` param. In this example we refetch the stream any time the userId changes. Without using the key param, even though we pass the `state.userId` into our function, it will only ever run the fetch for the userId that was part of the state when the stream was first added to the `evaluate` output.
```kotlin
if (state.isUserFetchEnabled && state.userId != null) {
    val fetchUserStream = RxStream.fromObservable(key = state.userId) { repository.fetchUser(state.userId) }
    fetchUserStream.onEvent { userResult ->
        // Do something with the result
    }
}
```

rjaylwar: I also think that because `Compose` has a declarative api we might be able to find some parallels like `LaunchedEffect`, `DisposableEffect` or the `remember(...) { ... }` function and make note of them. That way people who are already familiar with Composes mental model will have a head start in understanding Formulas.

PR #240, opened 2021-09-16
Coroutine flow support

Laimiux:
## What
Most of the logic is coming from #233 with some minor modifications.

- New `formula-coroutines` module.
- One example: `stopwatch-coroutines` sample.


formula-coroutines/src/test/java/com/instacart/formula/test/CoroutineTestExtensions.kt
```diff
@@ -0,0 +1,28 @@
+package com.instacart.formula.test
```
Laimiux, line 1: I placed coroutine test logic in `formula-coroutines/test` folder for now. It's a pending item to figure out how to best organize this and I don't want this to block merging these changes.

carrotkite:
### JaCoCo Code Coverage 76.48% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/coroutines/FlowFormula` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/coroutines/FlowRuntime` | 100% | 0% | :white_check_mark: |


</p>

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowRuntime.kt
```diff
@@ -0,0 +1,42 @@
...
+        input: Flow<Input>,
+        formula: IFormula<Input, Output>
+    ): Flow<Output> {
+        val threadChecker = ThreadChecker()
+        return callbackFlow<Output> {
+            threadChecker.check("Need to subscribe on main thread.")
+
+            var runtime = FormulaRuntime(threadChecker, formula, ::trySendBlocking, channel::close)
+
+            input.collect { input ->
```
Laimiux, line 24: TODO: Using `collect` here is a bug, need to fix it

emmano: would you mind expanding a  bit on why using `collect` is a bug here?

emmano: Is it because currently there is no way to cancel it? Meaning there is no job instance that would allow us to dispose of it?

Laimiux: Not exactly, it's actually hard to see it if you aren't very familiar with coroutines (I'm still pretty new so this took me a bit to figure out). I found this bug as I was adding more tests to the coroutine runtime. The issue came up when I added a test for formula termination. 

The problem here is that `collect` is a suspending operation and it will block execution until the flow is exhausted. When you pass `MutableSharedFlow`, it will never exhaust itself so nothing below `input.collect{}` will run including `awaitClose`.

To fix this, I replaced `collect` with
```kotlin
input.onEach {}.launchIn(this)
awaitClose {}
```

emmano: Thanks for the explanation. You are exactly right, those hot Flows never finish. I was going to propose what you suggest as a solution as well.

The question I have about that is, do we need to call job.cancel() on the job returned by launchIn() inside awaitClose inside the if?

Laimiux: I don't think we need to do that as we pass our `scope` to `launchIn(this)` so it should be canceled once the `callbackFlow` is canceled.

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowRuntime.kt
```diff
@@ -0,0 +1,43 @@
...
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.channels.trySendBlocking
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+
+object FlowRuntime {
```
mattkranzler5, line 15: Do we want to add `@ExperimentalCoroutinesApi` here or opt in at the module level?

mattkranzler5 (approved):
Looks good! Only question is how to handle the experimental calls.

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowRuntime.kt
Laimiux: Good question - I will mark this method with this annotation so that the consumers are aware when using the API.

formula-coroutines/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+}
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+dependencies {
+    implementation libraries.kotlin
+    implementation libraries.coroutines
```
Laimiux, line 13: Likely this should be exposed using `api`

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowRuntimeExtensions.kt
```diff
@@ -0,0 +1,25 @@
+package com.instacart.formula.coroutines
+
+import com.instacart.formula.IFormula
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+
+@ExperimentalCoroutinesApi
+fun <Output : Any> IFormula<Unit, Output>.toFlow(): Flow<Output> {
```
colinrtwhite, line 9: If these are experimental I'd probably create your own experimental annotation instead of reusing `ExperimentalCoroutinesApi`. `ExperimentalFormulaApi` would follow the same convention.

mattkranzler5: I don't think they are experimental (just implementation details) so I think the (more) correct approach would be to opt in and hide it from the consumer.

Laimiux: @mattkranzler5 Hmm, since Formula depends on experimental (non-stable) APIs of coroutines, there might be an issue where a consumer updates coroutines to a version where these non-stable APIs have changed. This could cause a compilation issue or maybe even a runtime issue in Formula. If we opt-in, we are hiding this information from the consumer.

@colinrtwhite I don't have a good understanding yet of how these experimental annotations should be used. The public API of the `toFlow()` is not really experimental, it just relies on experimental APIs within its implementation. Is the best practice to create a custom annotation in those cases?

colinrtwhite: If it's only using experimental coroutines APIs then I'd vote to use `@OptIn(ExperimentalCoroutinesApi::class)` as well. In practice, the experimental coroutines APIs go through a deprecation cycle before being removed so it's relatively safe to hide their usage (using opt in) as an implementation detail.

Laimiux: Cool, will change it to `@OptIn`.

PR #243, opened 2021-09-21
Moving tests into FormulaRuntimeTest.

Laimiux:
## What
Continuing to move test into `FormulaRuntimeTest`. This helps us to consolidate duplicate tests + run tests for both Coroutine and RxJava runtimes.


formula/src/test/java/com/instacart/formula/subjects/StreamEventFormula.kt
```diff
@@ -0,0 +1,4 @@
+package com.instacart.formula.subjects
+
+class StreamEventFormula {
+}
```
Jawnnypoo, line 4: Unused?

Jawnnypoo (approved):

formula/src/test/java/com/instacart/formula/subjects/StreamEventFormula.kt
Laimiux: Good catch, removed

PR #229, opened 2021-09-10
Rename Stream to Action.

Laimiux:
I want to rename `Stream` to `Action` 
- I think name is more fitting as it represents a deferred action that is executed and managed by Formula runtime. 
- Renaming `updates` to `actions` will keep the naming more consistent
- `Action.onInit` and similar methods seems better named than `Stream.onInit`

| Old      | New |
| -------------- | ----------------------- |
| `Stream`         | `Action`       | 
| `RxStream`     | `RxAction`   | 
| `FlowStream` | `FlowAction` |
| `BoundedStream`         | `BoundedAction`           | 
| `Evaluation.updates`        | `Evaluation.actions` |

PR #246, opened 2021-09-30
Rename Update to BoundStream.

Laimiux:
## What
Renaming `Update` to `BoundStream`. There are thoughts about renaming it to `BoundAction` but for time being as #229 is not finalized, I want to get rid of `Update` type here and keep it more closely associated with `Stream` object.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/Evaluation.kt
+++ b/formula/src/main/java/com/instacart/formula/Evaluation.kt
@@ -10,5 +10,5 @@ package com.instacart.formula
  */
 data class Evaluation<out Output>(
     val output: Output,
-    val updates: List<Update<*>> = emptyList()
+    val updates: List<BoundStream<*>> = emptyList()
 )
--- a/formula/src/main/java/com/instacart/formula/BoundStream.kt
+++ b/formula/src/main/java/com/instacart/formula/BoundStream.kt
@@ -1,6 +1,6 @@
 package com.instacart.formula
 
-class Update<Message>(
+class BoundStream<Message>(
     val key: Any,
     val stream: Stream<Message>,
     initial: (Message) -> Unit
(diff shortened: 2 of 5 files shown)


formula/src/main/java/com/instacart/formula/Evaluation.kt
```diff
@@ -10,5 +10,5 @@ package com.instacart.formula
  */
 data class Evaluation<out Output>(
     val output: Output,
-    val updates: List<Update<*>> = emptyList()
+    val updates: List<BoundStream<*>> = emptyList()
```
Laimiux, line 13: Will rename this and add deprecated logic here once #229 is finalized.

carrotkite:
### JaCoCo Code Coverage 76.48% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/internal/UpdateManager` | 94% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/FormulaContextImpl` | 50% | 0% | :white_check_mark: |
| `com/instacart/formula/FormulaContext` | 7% | 0% | :white_check_mark: |
| `com/instacart/formula/Evaluation` | 100% | 0% | :white_check_mark: |


</p>

Jawnnypoo (approved):
LGTM, can you explain more what "Bound" means in this context?

Laimiux:
> LGTM, can you explain more what "Bound" means in this context?

`Stream` only defines an action (that can produce 0..n events) - it doesn't have a listener/callback associated with it. We need a type that combines `Stream` and listener/callback `(Event) -> Unit` which handles events emitted by the `Stream`. Initially, we called this type `Update`, but it doesn't seem very fitting so I renamed it to `BoundStream` (stream bound to a listener).

PR #247, opened 2021-10-01
Consolidating Transition type.

Laimiux:
## What
Added new `Update<State, Event>` type which is a function that creates the `Transition` object. Maybe those two types should merge into `Update` and `Update.Result` or `Transition` and `Transition.Result` ?

### DIFF
--- a/formula/src/main/java/com/instacart/formula/FormulaContext.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaContext.kt
@@ -3,6 +3,7 @@ package com.instacart.formula
 import com.instacart.formula.internal.JoinedKey
 import com.instacart.formula.internal.ScopedCallbacks
 import com.instacart.formula.internal.TransitionDispatcher
+import com.instacart.formula.internal.UnitCallback
 
 /**
  * Provides functionality within [evaluate][Formula.evaluate] function to [compose][child]
@@ -10,7 +11,7 @@ import com.instacart.formula.internal.TransitionDispatcher
  * to arbitrary asynchronous events.
  */
 abstract class FormulaContext<State> internal constructor(
-    @PublishedApi internal val callbacks: ScopedCallbacks,
+    @PublishedApi internal val callbacks: ScopedCallbacks<State>,
     @PublishedApi internal val transitionDispatcher: TransitionDispatcher<State>,
 ) {
 
@@ -22,9 +23,7 @@ abstract class FormulaContext<State> internal constructor(
     fun <Event> onEvent(
         transition: UpdateType<State, Event>,
     ): Listener<Event> {
-        val reference = callbacks.initOrFindEventCallback<Event>(transition::class)
-        reference.delegate = transitionDispatcher.toCallback(transition)
-        return reference
+        return onEvent(key = transition::class, transition)
     }
 
     /**
@@ -37,7 +36,8 @@ abstract class FormulaContext<State> internal constructor(
         transition: UpdateType<State, Event>,
     ): Listener<Event> {
         val callback = callbacks.initOrFindEventCallback<Event>(key)
-        callback.delegate = transitionDispatcher.toCallback(transition)
+        callback.transitionDispatcher = transitionDispatcher
+        callback.transition = transition
         return callback
     }
 
@@ -47,9 +47,8 @@ abstract class FormulaContext<State> internal constructor(
      * It uses inlined callback anonymous class for type.
      */
     fun callback(transition: UpdateType<State, Unit>): () -> Unit {
-        val reference = callbacks.initOrFindCallback(transition::class)
-        reference.delegate = transitionDispatcher.toCallback(transition)
-        return reference
+        val event = onEvent(transition)
+        return UnitCallback(event)
     }
 
     /**
@@ -61,9 +60,8 @@ abstract class FormulaContext<State> internal constructor(
         key: Any,
         transition: UpdateType<State, Unit>
     ): () -> Unit {
-        val callback = callbacks.initOrFindCallback(key)
-        callback.delegate = transitionDispatcher.toCallback(transition)
-        return callback
+        val event = onEvent(key, transition)
+        return UnitCallback(event)
     }
 
     /**
--- a/formula/src/main/java/com/instacart/formula/FormulaContext.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaContext.kt
@@ -12,7 +12,7 @@ import com.instacart.formula.internal.UnitCallback
  */
 abstract class FormulaContext<State> internal constructor(
     @PublishedApi internal val callbacks: ScopedCallbacks<State>,
-    private val transitionDispatcher: TransitionDispatcher<State>,
+    internal val transitionDispatcher: TransitionDispatcher<State>,
 ) {
 
     /**
@@ -128,7 +128,7 @@ abstract class FormulaContext<State> internal constructor(
      * Provides methods to declare various events and effects.
      */
     class UpdateBuilder<State> internal constructor(
-        private val transitionDispatcher: TransitionDispatcher<State>
+        private val formulaContext: FormulaContext<State>,
     ) {
         internal val updates = mutableListOf<BoundStream<*>>()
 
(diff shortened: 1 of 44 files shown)


carrotkite:
### JaCoCo Code Coverage 79.03% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/internal/TransitionDispatcher` | 75% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/TransitionUtils` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/FormulaManagerImpl` | 88% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/ListenerImpl` | 75% | 0% | :white_check_mark: |
| `com/instacart/formula/FormulaContext` | 47% | 0% | :white_check_mark: |


</p>

docs/async_events.md
```diff
@@ -30,8 +30,8 @@ override fun evaluate(input: Input, state: State, context: FormulaContext): ...
 
 Formula uses a `Stream` interface to define an asynchronous event producers/sources.
 ```kotlin
-interface Stream<Message> {
-  fun start(send: (Message) -> Unit): Cancelable?
+interface Stream<Event> {
```
Laimiux, line 33: For consistency sake, removing all `Message` instances with `Event`

formula/src/test/java/com/instacart/formula/subjects/StartStopFormula.kt
```diff
@@ -40,13 +42,16 @@ class StartStopFormula(runtime: TestableRuntime) : Formula<Unit, State, Output>
...
-                    transition(state.copy(listenForEvents = false))
-                }
+                // We need to specify keys since `UpdateListenFlag` type is used two times.
+                startListening = context.onEvent("start", UpdateListenFlag(listen = true)),
+                stopListening = context.onEvent("stop", UpdateListenFlag(listen = false)),
             )
         )
     }
+
+    private class UpdateListenFlag(val listen: Boolean): Transition<State, Unit> {
```
Laimiux, line 52: Example of using separate type for `Transition`

formula/src/test/java/com/instacart/formula/subjects/StartStopFormula.kt
```diff
@@ -40,13 +42,16 @@ class StartStopFormula(runtime: TestableRuntime) : Formula<Unit, State, Output>
...
+        override fun TransitionContext<State>.toResult(event: Unit): Transition.Result<State> {
+            return transition(state.copy(listenForEvents = listen))
```
Laimiux, line 54: The `state` is provided by the `TransitionContext`. I'm thinking it would be nice to also provide the `Input`, however, this would most likely require adding the `Input` type to `FormulaContext`.

Laimiux: To support `Input` type and maybe to improve concepts, maybe `FormulaContext` should be renamed to `EvaluationContext<Input, State>` and we could simplify logic to 
```kotlin
interface Formula<Input, State, Output> {
    fun EvaluationContext<Input, Output>.evaluate(): Evaluation<Output> 
}
```

There is still a bit to consider here, but would be an interesting approach where you would get `input` and `state` from the `EvaluationContext`.

formula/src/test/java/com/instacart/formula/subjects/StartStopFormula.kt
mattkranzler5: Would be interesting to add transition specific unit tests.

formula/src/main/java/com/instacart/formula/Transition.kt
```diff
@@ -1,57 +1,51 @@
...
  */
-sealed class Transition<out State> {
-    companion object {
+fun interface Transition<State, in Event> {
+
+    /**
+     * Defines a transition result which can request a state change and/or some action
+     * to be performed.
+     */
+    sealed class Result<out State> {
```
johnwilde, line 13: I wonder if `Result` could cause confusion. Would `Kind` be better?

johnwilde (approved):

formula/src/test/java/com/instacart/formula/subjects/StartStopFormula.kt
Laimiux: Sorry, not sure I understand this fully. Could you explain a bit more?

formula/src/main/java/com/instacart/formula/Transition.kt
Laimiux: Could you explain the confusion with `Result` a bit more? Maybe we can find a different name, though I'm not sure I like `Kind` more. When I think of `Kind` (or `Type`), I usually think of some static enum and not something that contains dynamic data.

johnwilde: I thought confusion might come from thinking that a `Result` holds the result of a transition that _has_ occurred, whereas I think this is used to specify the "kind of transition" that _will_ occur.

Laimiux: What about `Transition.Event` since it will be sent to `FormulaManager`?

formula/src/test/java/com/instacart/formula/subjects/StartStopFormula.kt
mattkranzler5: I meant you could create unit tests specifically for transitions instead of for the whole formula if you create classes that encapsulate transitions.

Laimiux: Exactly, though from my experience, I would still strongly suggest testing the whole formula as those tests tend to be more robust.

Jawnnypoo (approved):

formula/src/main/java/com/instacart/formula/FormulaContext.kt
```diff
@@ -21,7 +21,7 @@ abstract class FormulaContext<State> internal constructor(
      * It uses [transition] type as key.
      */
     fun <Event> onEvent(
-        transition: Transition.Factory.(Event) -> Transition<State>,
+        transition: Transition<State, Event>,
```
FrancoisBlavoet, line 24: looks way more understandable than before 👍 

Transition.Factory.(Unit) -> Transition<State>
does not scan as something very intuitive

formula/src/main/java/com/instacart/formula/Transition.kt
```diff
@@ -1,79 +1,50 @@
 package com.instacart.formula
 
 /**
- * Defines an intent to transition by emitting a new [State] and optional [Effects].
+ * Transition is a function that is called when an [Event] happens and produces a [Result] which
```
FrancoisBlavoet, line 4: 👍

formula/src/main/java/com/instacart/formula/Transition.kt
FrancoisBlavoet: >holds the result of a transition that has occurred

My understanding is that it is correct though, Result holds the transformation which has been applied 

We could always use .Data is we wanted something completly neutral, but fwiw Result does not shock me

formula/src/test/java/com/instacart/formula/subjects/StartStopFormula.kt
FrancoisBlavoet: I like EvaluationContext, it better describes what this does 👍 

Due to how it is used on Android "Context" kind of sounds like some god object which isn't tied to this specific evaluation

I don't know if this is actually tripping people discovering formula though ?

FrancoisBlavoet (approved):

formula/src/main/java/com/instacart/formula/Transition.kt
alexanderbezverhni: I personally don't have a strong opinion, `Transition.Event` feels a bit better to me.

formula/src/main/java/com/instacart/formula/Transition.kt
```diff
@@ -1,79 +1,50 @@
...
+     * Defines a transition result which can request a state change and/or some action
+     * to be performed.
      */
-    object None : Transition<Nothing>() {
-        override val effects: Effects? = null
-    }
-
+    sealed class Result<out State> {
+        /**
+         * Stateful transition.
```
alexanderbezverhni, line 16: `Stateful transition result.`?

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -38,20 +38,20 @@ internal class FormulaManagerImpl<Input, State, Output>(
 
     private var childTransitionListener: TransitionListener? = null
 
-    private fun handleTransition(transition: Transition<State>) {
+    private fun handleTransition(result: Transition.Result<State>) {
```
alexanderbezverhni, line 41: `handleTransitionResult`?

alexanderbezverhni (approved):

PR #256, opened 2021-10-11
[Bugfix] Fix double key decoration. 

Laimiux:
## What
When `onEvent` was called from `StreamBuilder` it would create a double nested key:
```kotlin
JoinedKey(type, JoinedKey(Unit, type))
```

To avoid this, I've created an internal `eventListener` function.


Laimiux:
> makes sense
> 
> More generally (and outside of the scope of this PR) Could formula have some minified tests to detect future R8 issues early ?

Yeah, I've been trying to do that unsuccessfully. I think it's only possible with instrumentation tests and we don't have any in this repo.

PR #273, opened 2022-06-22
[Core] Add auto indexing of keys.

Laimiux:
## What
Instead of crashing when listener with duplicate key is created, let's auto-index it instead.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/internal/Listeners.kt
+++ b/formula/src/main/java/com/instacart/formula/internal/Listeners.kt
@@ -2,31 +2,29 @@ package com.instacart.formula.internal
 
 internal class Listeners {
     private var listeners: SingleRequestMap<Any, ListenerImpl<*, *, *>>? = null
+    private var indexes: MutableMap<Any, Int>? = null
 
     private fun duplicateKeyErrorMessage(key: Any): String {
-        if (key is String) {
-            // This indicates manual key creation.
-            return "Listener $key is already defined. Make sure your key is unique."
-        }
-        // This indicates automatic key generation
-        return "Listener $key is already defined. Are you calling it in a loop or reusing a method? You can wrap the call with FormulaContext.key"
+        return "Listener $key is already defined. Unexpected issue."
     }
 
     fun <Input, State, Event> initOrFindListener(key: Any): ListenerImpl<Input, State, Event> {
-        val listeners = listeners ?: run {
-            val initialized: SingleRequestMap<Any, ListenerImpl<*, *, *>> = mutableMapOf()
-            this.listeners = initialized
-            initialized
+        val currentHolder = listenerHolder<Input, State, Event>(key)
+        return if (currentHolder.requested) {
+            val index = nextIndex(key)
+            val indexedKey = IndexedKey(key, index)
+            initOrFindListener(indexedKey)
+        } else {
+            currentHolder
+                .requestAccess {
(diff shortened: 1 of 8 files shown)


formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
```diff
@@ -428,9 +430,21 @@ class FormulaRuntimeTest(val runtime: TestableRuntime, val name: String) {
     }
 
     @Test
-    fun `using listeners within another function crashes`() {
-        val result = Try { UsingCallbacksWithinAnotherFunction.test(runtime) }
-        assertThat(result.errorOrNull()?.cause).isInstanceOf(IllegalStateException::class.java)
```
Laimiux, line 433: Doesn't crash any more.

formula/src/main/java/com/instacart/formula/internal/Listeners.kt
```diff
@@ -2,31 +2,29 @@ package com.instacart.formula.internal
...
     fun <Input, State, Event> initOrFindCallback(key: Any): ListenerImpl<Input, State, Event> {
-        val listeners = listeners ?: run {
-            val initialized: SingleRequestMap<Any, ListenerImpl<*, *, *>> = mutableMapOf()
-            this.listeners = initialized
-            initialized
+        val currentHolder = listenerHolder<Input, State, Event>(key)
+        return if (currentHolder.requested) {
+            val index = nextIndex(key)
+            val indexedKey = IndexedKey(key, index)
+            initOrFindCallback(indexedKey)
```
alexanderbezverhni, line 16: Maybe add a check here that `key != indexedKey`, to avoid possible endless recursion?

formula/src/main/java/com/instacart/formula/internal/Listeners.kt
```diff
@@ -40,4 +38,30 @@ internal class Listeners {
...
+    /**
+     * Function which returns next index for a given key. It will
+     * mutate the [indexes] map.
+     */
+    private fun nextIndex(key: Any): Int {
+        val indexes = indexes ?: run {
+            val initialized = mutableMapOf<Any, Int>()
+            this.indexes = initialized
+            initialized
+        }
```
alexanderbezverhni, line 51: Maybe I'm missing something, but looks like this could be simplified to `val indexes = this.indexes ?: mutableMapOf<Any, Int>()`?

formula/src/main/java/com/instacart/formula/internal/Listeners.kt
Laimiux: I might be missing something, but I don't think it's possible for them to ever be equal.

formula/src/main/java/com/instacart/formula/internal/Listeners.kt
Laimiux: We need to assign `indexes` variable after we initialize the map 
```kotlin
this.indexes = initialized
```

Laimiux: We are using lazy instantiation to avoid unnecessary allocations internally.

alexanderbezverhni (approved):
LGTM

formula/src/main/java/com/instacart/formula/internal/Listeners.kt
Laimiux: Added an extra check

PR #274, opened 2022-06-23
Github Actions.

Laimiux:


.github/workflows/build-workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Run Formula Android Unit Tests
+        run: ./gradlew :formula-android:testRelease
+      - name: Run Formula Android Instrumentation Tests
+        run: ./gradlew :formula-android-tests:testRelease
+      - name: Run Formula Lint Tests
+        run: ./gradlew :formula-lint:test
+      - name: Generate Jacoco Report
+        run: ./gradlew jacocoTestReportMerged
+      - name: Save Jacoco Report
+        if: always()
```
alexanderbezverhni, line 27: Is this a legit syntax? I saw `${{ always() }}` in GitHub samples. Also, was the intention here to always run this job, even if some of the previous jobs failed?

Laimiux: Reading about it `always()` works as well.

alexanderbezverhni (approved):

PR #275, opened 2022-06-23
Enable context.key to handle all use cases.

Laimiux:
## What
Adding full support for `context.key`
- Child formulas now respect this value
- Nested scopes are now supported

Child formulas now respect `context.key`
```kotlin
context.key(item.id) {
  val imageOutput = context.child(itemImageFormula, item.imageUrl)
}
```

Nested `context.key` calls are now supported
```kotlin
context.key("first") {
  context.key("second") {
     // etc
  }
}
```


formula/src/main/java/com/instacart/formula/internal/SnapshotImpl.kt
```diff
@@ -43,10 +38,36 @@ internal class SnapshotImpl<out Input, State> internal constructor(
...
+    override fun endScope() {
+        if (scopeKey == null) {
+            throw IllegalStateException("Cannot end root scope.")
+        }
+
+        val lastKey = (scopeKey as? JoinedKey)?.left
+        scopeKey = lastKey
+    }
+
+    override fun ensureKeyIsScoped(type: KClass<*>, key: Any?): Any {
```
johnwilde, line 62: `createScopedKey`?

johnwilde (approved):
Should example be added to demonstrate new functionality?

PR #276, opened 2022-06-24
Create ChildrenManager.

Laimiux:
## What
Refactoring internals a little bit (moving all children logic into a separate manager similarly to `Listeners` and `ActionManager`)


formula/src/main/java/com/instacart/formula/internal/ChildrenManager.kt
```diff
@@ -0,0 +1,84 @@
...
+
+import com.instacart.formula.IFormula
+
+/**
+ * Keeps track of child formula managers.
+ */
+internal class ChildrenManager(
+    private val childTransitionListener: TransitionListener,
+) {
+    private var children: SingleRequestMap<Any, FormulaManager<*, *>>? = null
```
Laimiux, line 11: Logic copied over from `FormulaManagerImpl`

johnwilde (approved):

PR #277, opened 2022-06-24
Remove not needed key allocation.

Laimiux:


formula/src/main/java/com/instacart/formula/internal/SnapshotImpl.kt
```diff
@@ -60,6 +60,11 @@ internal class SnapshotImpl<out Input, State> internal constructor(
     }
 
     override fun createScopedKey(type: KClass<*>, key: Any?): Any {
+        if (scopeKey == null && key == null) {
+            // No need to allocate a new object, just use type as key.
+            return type
```
alexanderbezverhni, line 65: Probably worth adding a test for this?

Laimiux: There really isn't a great way to test this independently. However, there are a ton of tests within `FormulaRuntimeTest` which would capture if something would break with key management.

PR #278, opened 2022-06-27
Upgrade dependencies and API versions

chao2zhang:


chao2zhang:
@Laimiux When using includedBuild, AGP will check internally to see if the includedBuild's AGP is compatible with the consuming project.

Exception thrown by 
- https://android.googlesource.com/platform/tools/base/+/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/utils/agpVersionChecker.kt#84
- https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:android/src/com/android/tools/idea/gradle/project/upgrade/AgpVersionChecker.kt;l=25?q=agpVersionChecker

PR #300, opened 2023-06-27
Update AGP 8

Jawnnypoo:
Did this when trying to explore some issues with R8 full mode. 

- Updates to use Java 17 and target Java 11 (required for AGP 8)
- Update to AGP 8.0
- Update to Gradle 8.1.1
- Update RxJava and Kotlin versions

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ apply plugin: "com.github.ben-manes.versions"
 
 buildscript {
     ext {
-        kotlinVersion = '1.5.21'
+        kotlinVersion = '1.8.22'
 
         versions = [:]
         libraries = [:]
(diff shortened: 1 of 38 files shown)


build.gradle
```diff
@@ -21,11 +21,13 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.0.2'
+        classpath 'com.android.tools.build:gradle:8.0.2'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
-        classpath "com.vanniktech:gradle-android-junit-jacoco-plugin:0.16.0"
+        // We need to fork the plugin to fix an incorrect usage of tasks.withType(Test)
+        // https://github.com/vanniktech/gradle-android-junit-jacoco-plugin/compare/master...Laimiux:gradle-android-junit-jacoco-plugin:laimonas/gradle-8?expand=1
+        classpath("com.github.Jawnnypoo:gradle-android-junit-jacoco-plugin:61a9cc0129")
```
Jawnnypoo, line 28: Fork in order to support publishing on Jitpack. Hoping that we can see https://github.com/vanniktech/gradle-android-junit-jacoco-plugin/pull/219 get merged and won't have to worry about this anymore.

formula-android/src/main/java/com/instacart/formula/android/FeatureView.kt
```diff
@@ -16,12 +16,12 @@ import io.reactivex.rxjava3.core.Observable
  * @param bind A bind function connects state observable to the view rendering.
  * @param lifecycleCallbacks Optional lifecycle callbacks if you need to know the Fragment state.
  */
-class FeatureView<RenderModel>(
+class FeatureView<RenderModel : Any>(
```
Jawnnypoo, line 19: It would not compile without adding this in various places, unfortunately. I believe this stems from RxJava (hence why I updated the dependency) but I am not 100% sure.

build.gradle
Laimiux: I don't think we publish to `Jitpack` anymore

Laimiux:
Really awesome, I was thinking that we need to update Gradle. One thing that constantly breaks with Gradle updates is the maven publishing plugin. Could you verify that publish to artifactory and to maven still run correctly?

build.gradle
Jawnnypoo: Sorry yeah, that was confusing wording. I forked your fork of `gradle-android-junit-jacoco-plugin` so that I could change configuration of that lib to publish it to Jitpack, since it was configured to publish to IC's internal artifactory server, and was failing to build on Jitpack.

Jawnnypoo:
Covered by https://github.com/instacart/formula/pull/324

PR #301, opened 2023-06-28
Rewriting internals for better performance

Laimiux:
Introducing new formula internals which optimize formula action execution.

When using old internals, Formula has to re-evaluate after each state change starting from the top/parent formula all the way down to the child formula that had the state change. For child formulas not in the direct path, they likely won’t have to re-evaluate as long as parent input matches the last one used.

The new formula internals optimize the scenario when during formula evaluation we start an action that emits a state change. Given we are already within the evaluation loop and haven’t returned an output to the parent, we can re-evaluate inline without having to trigger full chain re-evaluation. This enables cutting a bunch of full re-evaluations. This optimization required us to move action start/stop to the evaluation.

Overall, it seems new internals perform about 27% better (and reduce about 80% full re-evaluations).

**Examples of storefront page load as end metric**
| Old | New |
| ---- | --- |
| storefront - formula summary - 369 - total: 370 ms | storefront - formula summary - 73 - total: 270 ms |
| storefront - formula summary - 361 - total: 374 ms | storefront - formula summary - 75 - total: 282 ms |
| storefront - formula summary - 361 - total: 356 ms | storefront - formula summary - 73 - total: 264 ms |
| storefront - formula summary - 359 - total: 450 ms | storefront - formula summary - 77 - total: 308 ms |
| storefront - formula summary - 367 - total: 378 ms | storefront - formula summary - 79 - total: 276 ms |

On average, new internals performs about 100 ms better with 280 ms / 75 runs versus 385 ms / 363 runs.


formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -69,47 +69,60 @@ class FormulaRuntime<Input : Any, Output : Any>(
...
+        val manager = checkNotNull(manager)
+        val currentInput = checkNotNull(input)
+
+        if (evaluate && !manager.terminated) {
+            isEvaluating = true
+            evaluationPhase(manager, currentInput)
+            isEvaluating = false
+        }
 
+        if (evaluate || effectQueue.isNotEmpty()) {
```
chao2zhang, line 116: `executionRequested` does not seem to be used anymore

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -39,12 +40,30 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
+        if (terminated) {
+            event.dispatch()
+        } else if (isEvaluating) {
+            eventQueue.addLast(event)
+        } else {
+            delegate.onPostponedTransition(event)
+        }
+    }
+
+    fun hasPendingTransitions(): Boolean {
```
chao2zhang, line 57: Just to confirm that this is the critical piece of optimization in this PR.

chao2zhang (approved):

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -75,7 +94,39 @@ internal class FormulaManagerImpl<Input, State, Output>(
      */
     override fun evaluate(input: Input): Evaluation<Output> {
         // TODO: assert main thread.
-        val transitionID = transitionID
+
+        var result: Evaluation<Output>? = null
+        isEvaluating = true
+        while (result == null) {
+            val lastFrame = frame
+            if (lastFrame != null && executeTransitions(lastFrame.transitionID)) {
```
jhowens89, line 102: We need to update the documentation for this class with the way we are now handling transitions here

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -125,25 +175,39 @@ internal class FormulaManagerImpl<Input, State, Output>(
         }
 
         val frame = Frame(snapshot, result, transitionID)
-        actionManager.onNewFrame(frame.evaluation.actions)
         this.frame = frame
 
+        actionManager.onNewFrame(frame.evaluation.actions)
```
jhowens89, line 180: What are the ramifications and/or intentions of moving this line? There seems to be a number of affected codepaths in evaluation if `actionManager.onNewFrame` caused re-evaluations.

formula/src/main/java/com/instacart/formula/internal/ManagerDelegate.kt
```diff
@@ -9,4 +9,8 @@ interface ManagerDelegate {
         result: Transition.Result<*>,
         evaluate: Boolean,
     )
+
+    fun onPostponedTransition(
```
jhowens89, line 13: You've kinda alluded to it in the PR description, but can we document onPostponedTransition vs onTransition?

formula/src/main/java/com/instacart/formula/Inspector.kt
```diff
@@ -61,6 +62,11 @@ interface Inspector {
      */
     fun onActionFinished(formulaType: KClass<*>, action: DeferredAction<*>) = Unit
 
+    /**
+     * Called when transition [Effects] are executed
+     */
+    fun onEffectExecuted() = Unit
```
jhowens89, line 68: This callback would probably be much more useful if we passed the effect here, right? Wouldn't it typically be an anonymous class hinting at where the execution point is?

Laimiux: I've removed this for now, will add it in another PR

formula/src/main/java/com/instacart/formula/internal/ManagerDelegate.kt
Laimiux: I've refactored `ManagerDelegate` to contain only `onUpdate`

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
Laimiux: No ramifications, it doesn't perform any side-effecting logic. It only invalidates internal ActionManager state which will be picked up in `FormulaManagerImpl.executeUpdates`

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
Laimiux: Removed

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
Laimiux: Yep, this is checked within `executeUpdates` block to know if we need to go back to `executeTransitions`

formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
```diff
@@ -316,7 +315,7 @@ class FormulaRuntimeTest(val runtime: TestableRuntime, val name: String) {
             .output { assertThat(child.state).isEqualTo(3) }
 
         inspector.assertRunCount(1)
-        inspector.assertEvaluationCount(HasChildFormula::class, 8)
+        inspector.assertEvaluationCount(HasChildFormula::class, 5)
```
Laimiux, line 318: The changes in this test class display the improvements in the number of evaluations.

formula/src/main/java/com/instacart/formula/internal/DeferredTransition.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.instacart.formula.Transition
+
+/**
+ * A deferred transition contains an [event] and a related [transition] which can
+ * be executed using the [execute] function. If the formula is ready, it will be
+ * executed immediately. Otherwise, it will be added to the queue and executed
+ * when formula is ready. If the formula is re-evaluated and [listener] is disabled
+ * before the transition is executed, this transition will be ignored.
+ */
+class DeferredTransition<Input, State, EventT> internal constructor(
```
Laimiux, line 12: Explicit type for deferred transitions.

formula/src/main/java/com/instacart/formula/internal/FormulaManager.kt
```diff
@@ -11,16 +11,13 @@ interface FormulaManager<Input, Output> {
...
-     * Called after [evaluate] to update child formulas and actions.
+     * Calls [Formula.evaluate] and after evaluation, it will process the new state
+     * by cleaning up detached child formulas, terminating old actions, and then starting
+     * new ones. If at any given point there is a state change, it will run [Formula.evaluate].
      *
-     * @return True if transition happened while performing this.
+     * It will handle formula state changes internally and return the last [Output].
      */
-    fun executeUpdates(): Boolean
+    fun run(input: Input): Evaluation<Output>
```
Laimiux, line 20: Consolidated `evaluate` and `executeUpdates` into a single `run` function.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -1,21 +1,21 @@
...
 
 /**
- * Handles formula and its children state processing.
- *
- * Order of processing:
- * 1. Evaluate
- * 2. Disable old event listeners
- * 3. Terminate removed children
- * 4. Prepare parent and alive children for updates.
+ * Responsible for keeping track of formula's state, running actions, and child formulas. The
```
Laimiux, line 14: Updated documentation

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -39,30 +39,53 @@ internal class FormulaManagerImpl<Input, State, Output>(
 
     var transitionID: Long = 0
     var terminated = false
+    private var isRunning: Boolean = false
```
Laimiux, line 42: We keep track if we are within the `run` function. If so, we can perform a bunch of logic inline without the need for a full re-run.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -39,30 +39,53 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
 
-    fun hasTransitioned(id: Long): Boolean {
+    private val transitionQueue = LinkedList<DeferredTransition<*, *, *>>()
```
Laimiux, line 44: Pending transition queue that will be executed after the latest evaluate.

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -128,53 +123,13 @@ class FormulaRuntime<Input : Any, Output : Any>(
...
             } finally {
                 manager.setValidationRun(false)
             }
         }
     }
 
-    /**
-     * Executes operations containing side-effects such as starting/terminating streams.
-     */
-    private fun executionPhase(manager: FormulaManagerImpl<Input, *, Output>) {
```
Laimiux, line 141: Side-effects are now executed by `FormulaManagerImpl`

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -128,53 +123,13 @@ class FormulaRuntime<Input : Any, Output : Any>(
...
-                continue
-            }
-        }
-        isExecuting = false
-    }
-
-    /**
-     * Executes effects from the [effectQueue].
-     */
-    private fun executeEffects(manager: FormulaManagerImpl<*, *, *>, transitionId: Long): Boolean {
```
Laimiux, line 164: Side-effects are now executed by `FormulaManagerImpl`

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -128,53 +123,13 @@ class FormulaRuntime<Input : Any, Output : Any>(
...
-     * Executes operations containing side-effects such as starting/terminating streams.
-     */
-    private fun executionPhase(manager: FormulaManagerImpl<Input, *, Output>) {
-        isExecuting = true
-        while (executionRequested) {
-            executionRequested = false
-
-            val transitionId = manager.transitionID
-            if (!manager.terminated) {
-                if (manager.executeUpdates()) {
```
Laimiux, line 148: Updates are now executed by `FormulaManagerImpl`

formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
```diff
@@ -1141,7 +1140,7 @@ class FormulaRuntimeTest(val runtime: TestableRuntime, val name: String) {
             assertThat(this).isEqualTo(250)
         }
         inspector.assertRunCount(1)
-        inspector.assertEvaluationCount(31625)
+        inspector.assertEvaluationCount(500)
```
chao2zhang, line 1143: 🔥

formula/src/main/java/com/instacart/formula/internal/ManagerDelegate.kt
```diff
@@ -1,12 +1,20 @@
...
-    fun onTransition(
-        formulaType: KClass<*>,
-        result: Transition.Result<*>,
-        evaluate: Boolean,
-    )
+    /**
+     * When [FormulaManagerImpl] is not currently running and needs to be evaluated again, it
+     * will request the parent delegate to trigger a new evaluation run.
+     */
+    fun requestEvaluation()
```
jhowens89, line 13: requestEvaluation seems somewhat of subset of `onPendingTransition(transition: DeferredTransition<*, *, *>)`. Can you help me understand the difference?

jhowens89 (approved):
Can't overstate the improvement the greater documentation coverage is. There are a lot of optimizations (and good reasons for them) in Formula and they can be painful to follow without some explanation. Obviously the new optimizations are awesome, but this PR adds a lot outside of that.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -40,42 +40,119 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
+    fun isEvaluationNeeded(id: Long): Boolean {
         return transitionID != id
     }
 
+    fun isTerminated(): Boolean {
+        return terminated
+    }
+
     fun handleTransitionResult(result: Transition.Result<State>) {
+        val frame = checkNotNull(frame) {
```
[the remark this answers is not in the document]
Laimiux: Renamed to `lastFrame` and moved it up to where it is used.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -40,42 +40,119 @@ internal class FormulaManagerImpl<Input, State, Output>(
     var transitionID: Long = 0
```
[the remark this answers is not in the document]
Laimiux: These are good points and questions, I will touch on each individual point below. 

I update documentation around `transitionId` (probably better name would be `evaluationId`). 

> Does frame actually have any use for the transitionID member variable? Could we just maintain the single FormulaManagerImpl.transitionId?

Frame is associated with a specific `transitionId` value which allows us to be able to verify if it is still valid or if a new one needs to be created. Without placing it on `SnapshotImpl` or `Frame`, it's hard to know if we are in an out-dated state.

> Does transitionID have any use as a Long? Would it be more understandable as a Boolean? If a boolean is an appropriate use case I wonder if passing it around as argument is necessary.

We might be able to use a boolean, but I found the ID pattern easier to manage. With a boolean, you need to toggle it off in the right place, and it might lead to race conditions if a new event arrives. For example, if someone accidentally calls a listener within `Formula.evaluate`, `transitionId` makes it much easier to catch. In general, we might be able to use `boolean`, but I'm not sure if it will lead to a simpler system.

> I can imagine up scenarios where this might be necessary but not sure if real use cases. Hence some docs could help.

Agreed, I will update the documentation and will rename `transitionId` with `evaluationId` (as I think that better represents why we have it).

> Tangentially related: Should Frame be an inner class or private to the file? It appears to only be used by FormulaManagerImpl. You don't seem to have those elsewhere, but IMO it would help. Formula has many moving parts w/ different roles and if I can easily see that a class is only used locally it helps make digesting the control flow easier.

Yeah, likely it could be a private class within `FormulaManagerImpl`

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -180,11 +247,81 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
     }
 
-    override fun onTransition(formulaType: KClass<*>, result: Transition.Result<*>, evaluate: Boolean) {
-        if (evaluate) {
-            transitionID += 1
+    override fun onPendingTransition(transition: DeferredTransition<*, *, *>) {
+        if (terminated) {
+            transition.execute()
+        } else if (isRunning) {
+            transitionQueue.addLast(transition)
```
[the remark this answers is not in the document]
Laimiux: Yep, exactly. It specifically handles the cases when a child listener is triggered by another running formula. Some scenarios where this happens: 
- One formula triggers a global state change or relay that other formulas are listening to. 
- Parent calls child listener while running directly. For example, on action start, we want to notify the child formula of something
- Parent passes child listener to another child and that child calls it while running.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -40,42 +40,119 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
-        if (evaluate || result.effects != null) {
-            delegate.onTransition(loggingType, result, evaluate)
+        if (isRunning) {
+            if (effects != null) {
+                transitionEffectQueue.addLast(effects)
+            }
+        } else {
+            if (isEvaluationNeeded(frame.transitionID)) {
+                if (effects != null) {
+                    transitionEffectQueue.addLast(effects)
```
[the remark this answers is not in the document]
Laimiux: Yep, seems reasonable to clear the `transitionEffectQueue`, will write test first to simulate this scenario.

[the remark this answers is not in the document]
Laimiux: Actually, I'm not sure how to replicate this scenario in a test. I've attempted to do so with `ExecuteEffectAfterChildTerminatesRobot`, but wasn't able to. It's not easy to force the parent to detach the child while re-evaluating due to child's state change.

formula/src/main/java/com/instacart/formula/internal/ManagerDelegate.kt
Laimiux: Good point, I was able to remove `onPendingTransition` from the interface and only keep `requestEvaluation`

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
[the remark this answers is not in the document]
Laimiux: Actually, removing `onPendingTransition` from the interface broke `ExecuteEffectAfterChildTerminatesRobot`, and adding the transition effect queue clean up within termination fixed it.

formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
```diff
@@ -301,7 +311,8 @@ class FormulaRuntimeTest(val runtime: TestableRuntime, val name: String) {
         runtime.test(formula, Unit, inspector)
             .output { assertThat(state).isEqualTo(3) }
 
-        inspector.assertRunCount(1)
```
Laimiux, line 304: We changed how runs are calculated so this value is still 4.

formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
```diff
@@ -315,7 +326,9 @@ class FormulaRuntimeTest(val runtime: TestableRuntime, val name: String) {
         runtime.test(formula, Unit, inspector)
             .output { assertThat(child.state).isEqualTo(3) }
 
-        inspector.assertRunCount(1)
```
Laimiux, line 318: It looks like we regressed, but we changed how each run is calculated.

PR #302, opened 2023-06-29
Adding more tests to cover action lifecycle.

Laimiux:
Adding more tests in preparation for #301

### DIFF
--- a/formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
+++ b/formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
@@ -6,6 +6,7 @@ import com.instacart.formula.actions.EmptyAction
 import com.instacart.formula.internal.TestInspector
 import com.instacart.formula.internal.Try
 import com.instacart.formula.rxjava3.RxAction
+import com.instacart.formula.subjects.ChildActionFiresParentEventOnStart
 import com.instacart.formula.subjects.ChildMessageNoParentStateChange
 import com.instacart.formula.subjects.ChildMessageTriggersEventTransitionInParent
 import com.instacart.formula.subjects.ChildMessageWithParentStateChange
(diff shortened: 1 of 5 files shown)


formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
```diff
@@ -280,6 +282,59 @@ class FormulaRuntimeTest(val runtime: TestableRuntime, val name: String) {
...
+            .output { assertThat(value).isEqualTo(6) }
+            .apply {
+                // Efficiently we emit only emit 2 values
+                assertThat(values()).hasSize(2)
+            }
+    }
+
+    @Test
+    fun `on action start child triggers action in a parallel child`() {
+        val events = listOf(Unit, Unit, Unit, Unit)
```
Jawnnypoo, line 313: you could do `List(4) { Unit }` here and pull 4 out into a val that they can all reference

formula/src/test/java/com/instacart/formula/internal/TestInspector.kt
```diff
@@ -20,14 +20,6 @@ class TestInspector : Inspector {
...
-    override fun onExecutionStarted() {
-        events.add("execution-started")
-    }
-
-    override fun onExecutionFinished() {
-        events.add("execution-finished")
-    }
-
     override fun onActionStarted(formulaType: KClass<*>, action: DeferredAction<*>) {
         events.add("action-started: ${formulaType.qualifiedName}")
```
Jawnnypoo, line 24: Do you expect that outside consumers will be reading these events? if so, maybe pulling `action-started` and others out into a const so that consumers can check for these values would be nice

Jawnnypoo (approved):

formula/src/test/java/com/instacart/formula/internal/TestInspector.kt
Laimiux: `TestInspector` is not part of the library, it's just within formula tests.

carrotkite:
### JaCoCo Code Coverage 78.82% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/FormulaRuntime` | 83% | 0% | :white_check_mark: |


</p>

formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
Laimiux: Will update in follow-up

PR #303, opened 2023-06-29
Remove global transitionID.

Laimiux:
To prepare for #301, I've removed global transition ID and instead each `FormulaManager` will maintain its own.


formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -75,6 +60,18 @@ class FormulaRuntime<Input : Any, Output : Any>(
         }
     }
 
+    override fun onTransition(formulaType: KClass<*>, result: Transition.Result<*>, evaluate: Boolean) {
```
Laimiux, line 63: Renamed `TransitionListener` to `ManagerDelegate`

formula/src/main/java/com/instacart/formula/Inspector.kt
```diff
@@ -88,8 +88,8 @@ interface Inspector {
      *
      * @param formulaType Formula type used to filter for specific events.
      * @param result Transition result used to check what kind of transition.
-     * @param requiresEvaluation Indicates if transition requires a new evaluation.
+     * @param evaluate Indicates if transition requires a new evaluation.
```
Laimiux, line 91: Simpler name

formula/src/main/java/com/instacart/formula/internal/FormulaManager.kt
```diff
@@ -13,34 +13,24 @@ interface FormulaManager<Input, Output> {
...
-        transitionId: TransitionId
-    ): Evaluation<Output>
-
-    /**
-     * Method updates the transition id of the existing evaluation. This
-     * method is called when there were no changes and evaluation does
-     * not need to run in this part of the formula tree.
-     */
-    fun updateTransitionId(transitionId: TransitionId)
+    fun evaluate(input: Input): Evaluation<Output>
```
Laimiux, line 16: Removing global `TransitionId` allows us to simplify quite a few functions and also allows us to remove `updateTransitionId` which would have to iterate across the whole formula tree.

carrotkite:
### JaCoCo Code Coverage 78.87% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/internal/SnapshotImpl` | 81% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/PendingFormulaManagerList` | 85% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/FormulaManagerImpl` | 85% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/ChildrenManager` | 83% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/Frame` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/ActionManager` | 87% | 0% | :white_check_mark: |
| `com/instacart/formula/FormulaRuntime` | 82% | 0% | :white_check_mark: |


</p>

formula/src/main/java/com/instacart/formula/Inspector.kt
Jawnnypoo: I guess we can rename without backwards compatibility since no releases were done between this class getting created and it releasing as a version you can depend on?

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -31,47 +32,50 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
 
+    fun hasTransitioned(id: Long): Boolean {
+        return transitionID != id
+    }
+
     fun handleTransitionResult(result: Transition.Result<State>) {
         if (terminated) {
             // State transitions are ignored, only side effects are passed up to be executed.
-            transitionListener.onTransitionResult(type, result, true)
+            delegate.onTransition(type, result, false)
```
Jawnnypoo, line 50: looking at this without digging in, seems suspicious that it went from true to false. Maybe we can add named arguments here since at least in GitHub, its hard to know what is what.

Jawnnypoo (approved):

formula/src/main/java/com/instacart/formula/Inspector.kt
Laimiux: Yep, and also we don't provide such strong backwards compatibility guarantees (as it's pre-1.0).

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
Laimiux: Added named parameter for this

PR #304, opened 2023-06-29
Change order of updates where each leaf executes first.

Laimiux:
The updated order of execution:
- Running child formulas executes updates
    - Execute child formula updates
        - And the cycle continues...
    - Terminate detached child formulas
    - Terminate old actions
    - Start new actions
- Terminate previously detached child formulas
- Terminate old actions
- Start new actions

This allows us to run execute all child formula updates in one go.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
@@ -136,15 +136,7 @@ class FormulaRuntime<Input : Any, Output : Any>(
 
             val transitionId = manager.transitionID
             if (!manager.terminated) {
-                if (manager.terminateDetachedChildren()) {
-                    continue
-                }
-
-                if (manager.terminateOldUpdates()) {
-                    continue
-                }
-
-                if (manager.startNewUpdates()) {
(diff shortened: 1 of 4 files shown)


carrotkite:
### JaCoCo Code Coverage 78.76% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/internal/FormulaManagerImpl` | 86% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/ChildrenManager` | 83% | 0% | :white_check_mark: |


</p>

Jawnnypoo (approved):
Makes sense. This wouldn't change anything unexpectedly downstream, right?

Laimiux:
> Makes sense. This wouldn't change anything unexpectedly downstream, right?

Well, it definitely could. After this change:
- Child will start its actions before the parent
- Child will terminate its actions before the parent
- Child will terminate its child formulas before the parent

Essentially, we are switching the model from where initially the top of the tree was going first to where the leaves of the tree go first. Technically this shouldn't really matter much to the consumer of the public APIs as they shouldn't rely on such nuances for their features. However, it definitely can uncover such dependencies.

PR #307, opened 2023-07-05
Improvements to formula inspector logging

jhowens89:
Two suggestions to improve the Inspector logging:
* Pass the actual Formula and not the IFormula::implementation. If use just the implementation we get a lot of opage StatelessFormula, ActionFormula, etc.
* Account for local action starting, stopping, and transition execution during the the FormulaManagerImpl::evaluation callback since we have moved that logic there. We've had some edge cases where these were surprisingly expensive and it can be hard to discern the source if it is only reflected in root's run start/end.


formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -115,38 +125,33 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
         return result
     }
 
+    private data class EvaluationResult<Output>(val evaluation: Evaluation<Output>, val skipped: Boolean)
+
     /**
      * Creates the current [Output] and prepares the next frame that will need to be processed.
      */
-    private fun evaluation(input: Input, transitionID: Long): Evaluation<Output> {
+    private fun evaluation(input: Input, transitionID: Long): EvaluationResult<Output> {
```
jhowens89, line 141: I'm pretty sure there's a cleaner way to represent this logic, but I'm not confident enough to make too many assumptions. I'm thinking maybe the transitionID could be checked  instead of porting EvaluationResult.skipped.

PR #310, opened 2023-07-10
[Bugfix] Fix old listener crash.

Laimiux:
## What

There was a bug in the runtime after the rewrite in which it was possible to get `Transition already happened` exception:
```
Caused by: java.lang.IllegalStateException: Transition already happened. This is using old event listener: 
com.instacart.formula.subjects.IndirectNestedCallbackCallRobot$Parent$evaluate$2$2@3878c24f & 1. 
Transition: 2 != 3
```

This happens in a parent that has more than one child formula.
- Child A formula has already run and is idle 
- Child B formula is running and has a side-effect that causes Child A state to change
- Parent formula evaluation doesn't re-evaluate and causes us to have outdated listeners.

The way to solve this is to have `isEvaluationNeeded` call in `postEvaluation` before processing anything. This will force to re-evaluate again.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
+++ b/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
@@ -279,6 +279,10 @@ internal class FormulaManagerImpl<Input, State, Output>(
      * @return True if we need to re-evaluate.
      */
     private fun postEvaluation(evaluationId: Long): Boolean {
+        if (isEvaluationNeeded(evaluationId)) {
+            return true
+        }
+
         if (handleTransitionQueue(evaluationId)) {
             return true
         }
--- a/formula/src/test/java/com/instacart/formula/subjects/MultiChildIndirectStateChangeRobot.kt
+++ b/formula/src/test/java/com/instacart/formula/subjects/MultiChildIndirectStateChangeRobot.kt
@@ -1,22 +1,24 @@
 package com.instacart.formula.subjects
 
-import com.instacart.formula.Action
 import com.instacart.formula.Evaluation
 import com.instacart.formula.Formula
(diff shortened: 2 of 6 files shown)


formula/src/test/java/com/instacart/formula/subjects/MultiChildIndirectStateChangeRobot.kt
```diff
@@ -0,0 +1,120 @@
+package com.instacart.formula.subjects
+
+import com.instacart.formula.Evaluation
+import com.instacart.formula.Formula
+import com.instacart.formula.Snapshot
+import com.instacart.formula.rxjava3.RxAction
+import com.instacart.formula.test.TestableRuntime
+import io.reactivex.rxjava3.core.Observable
+
+class MultiChildIndirectStateChangeRobot(runtime: TestableRuntime) {
```
Laimiux, line 10: I've added a couple of tests, but this is the one that actually crashed.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -279,6 +279,10 @@ internal class FormulaManagerImpl<Input, State, Output>(
      * @return True if we need to re-evaluate.
      */
     private fun postEvaluation(evaluationId: Long): Boolean {
+        if (isEvaluationNeeded(evaluationId)) {
```
Laimiux, line 282: This fixes an issue where a child state might have changed during evaluation and wasn't picked up by us. This ensures we re-evaluate again.

Jawnnypoo (approved):

carrotkite:
### JaCoCo Code Coverage 78.81% :white_check_mark:
|:---|:---:|:---:|:---:|


</p>

PR #314, opened 2023-11-30
[Formula Android] Track first render.

Laimiux:
## What

Track the duration from when a formula fragment is created and when the first render model is applied to the view.


formula-android/src/main/java/com/instacart/formula/android/FeatureView.kt
```diff
@@ -22,6 +22,7 @@ class FeatureView<RenderModel>(
     val lifecycleCallbacks: FragmentLifecycleCallback? = null,
 ) {
     class State<RenderModel>(
+        val initialized: Long,
```
alexanderbezverhni, line 25: nit: `initializedAt` or `initializedAtMillis`?

alexanderbezverhni (approved):
👍🏼

PR #317, opened 2023-12-07
Handle duplicate child key errors more gracefully.

Laimiux:
## What
Instead of crashing on duplicate child key, we now add an index and notify the global listener to fix the problem. This is better behavior in production than potentially breaking the whole screen.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/FormulaPlugins.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaPlugins.kt
@@ -18,4 +18,12 @@ object FormulaPlugins {
             else -> ListInspector(listOf(global, local))
         }
     }
+
+    fun onDuplicateChildKey(
+        parentFormulaType: Class<*>,
+        childFormulaType: Class<*>,
+        key: Any,
+    ) {
+       plugin?.onDuplicateChildKey(parentFormulaType, childFormulaType, key)
+    }
 }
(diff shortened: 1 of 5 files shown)


formula/src/main/java/com/instacart/formula/FormulaPlugins.kt
```diff
@@ -18,4 +18,12 @@ object FormulaPlugins {
             else -> ListInspector(listOf(global, local))
         }
     }
+
+    fun onDuplicateChildKey(
```
dennisdeng2002, line 22: Do we want to generalize this to support other formula errors in the future? So something like:
```
onError(error: FormulaError)
```

where `FormulaError` is a sealed interface

```
sealed interface FormulaError {
    data class DuplicateKeyError(
        val parentFormulaType: KClass<*>,
        val childFormulaType: KClass<*>,
        val key: Any,
    ) : FormulaError
}
```

Laimiux: Maybe in the future, but for now doesn't seem necessary

jhowens89 (approved):

carrotkite:
### JaCoCo Code Coverage 79.54% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/internal/ChildrenManager` | 80% | 0% | :white_check_mark: |
| `com/instacart/formula/FormulaPlugins` | 87% | 0% | :white_check_mark: |


</p>

formula/src/main/java/com/instacart/formula/FormulaPlugins.kt
Laimiux: Given it's pretty easy to migrate to a different design, I prefer waiting for more than one use case before designing a generic thing.

dennisdeng2002 (approved):

PR #332, opened 2024-01-22
Make formula internals thread safe.

Laimiux:
## What

Currently, Formula is single-threaded and usually bound to the main UI thread. This requires processing all events and running evaluations on that thread. This can be a performance limitation:
- Evaluations can get pretty expensive for nested and complicated features 
- There can be delays jumping from network/io threads to the main thread (during initial screen loads when there is a lot of contention)

While we have to run only one formula evaluation at a time, there is no reason why this has to be bound to a single specific thread. This will allow network and other async events to be processed on IO thread instead of having to jump to main thread. 

### Implementation details
To make Formula unbound to any thread, we need to make the internals thread safe. Since, we can only handle one evaluation at a time, we implement this via a non-blocking FIFO queue for updates. The simplified way it works is when an event arrives, it checks if another thread is already processing the queue and if so, it hands off this update to that thread. If not, it starts to process the queue.


formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowRuntime.kt
```diff
@@ -25,15 +26,17 @@ object FlowRuntime {
     ): Flow<Output> {
         val threadChecker = ThreadChecker(formula)
         return callbackFlow<Output> {
-            threadChecker.check("Need to subscribe on main thread.")
+//            threadChecker.check("Need to subscribe on main thread.")
```
Laimiux, line 29: TODO: figure out if we just remove this, or have an opt-out for this behavior.

formula-coroutines/src/main/java/com/instacart/formula/coroutines/FlowRuntime.kt
```diff
@@ -46,17 +49,21 @@ object FlowRuntime {
...
+                    runtime.onInput(input)
                 }
-                runtime.onInput(input)
             }.launchIn(this)
 
             awaitClose {
-                threadChecker.check("Need to unsubscribe on the main thread.")
-                runtime.terminate()
+//                threadChecker.check("Need to unsubscribe on the main thread.")
+                queue.postUpdate {
```
Laimiux, line 64: TODO: need to check for race conditions given termination is deferred and some other events could happen after it.

formula-test/src/main/java/com/instacart/formula/test/TestFormulaObserver.kt
```diff
@@ -6,7 +6,7 @@ class TestFormulaObserver<Input : Any, Output : Any, FormulaT : IFormula<Input,
     private val delegate: FormulaTestDelegate<Input, Output, FormulaT>,
 ) {
 
-    private var started: Boolean = false
+    @Volatile private var started: Boolean = false
```
Laimiux, line 9: TODO: given the multi-threaded nature, we will need to go over the internals and ensure that thread memory visibility is configured correctly.

formula/src/main/java/com/instacart/formula/internal/SynchronizedEventQueue.kt
```diff
@@ -0,0 +1,74 @@
...
+     * - Input change
+     * - Individual formula transitions
+     * - Termination
+     *
+     * Implementation works by having a concurrent queue and checking:
+     *  - If queue is idle, execute current update and try to process other queue entries
+     *  - If queue is running by the same thread, we execute current update and let other
+     *  updates be handled by existing processing loop.
+     *  - If queue is running by a different thread, add to the queue and see if we need to
+     *  take over the processing.
```
tripathiamritansh, line 27: something to consider is the cost of context switching if we rapidly switch for small evaluation tasks. It might make sense to batch things if this can be the case

Laimiux: Could you explain more about what you mean by context switching?

tripathiamritansh: Thread context switching is the process of saving the state of a currently running thread and then restoring the state of another thread, allowing it to resume execution. This essentially allows multiple threads to share a single CPU, giving the illusion that they are running concurrently. The cost is fairly low for IO threads but high for computation threads

Laimiux: We are trying to avoid thread-switching and enable running the formula on the thread that the event is produced on. Currently, all network calls switch to the main thread and then process the event on that thread. In the future, we will not switch to the main thread, but process the event first and only switch to the main thread when updating the UI.

carrotkite:
### JaCoCo Code Coverage 78.76% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/coroutines/FlowRuntime` | 40% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/SynchronizedEventQueue` | 83% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/ChildrenManager` | 75% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/ActionManager` | 86% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/ListenerImpl` | 75% | 0% | :white_check_mark: |
| `com/instacart/formula/rxjava3/RxJavaRuntime` | 80% | 0% | :white_check_mark: |
| `com/instacart/formula/FormulaRuntime` | 76% | 0% | :white_check_mark: |
| `com/instacart/formula/test/TestFormulaObserver` | 50% | 0% | :white_check_mark: |


</p>

formula/src/main/java/com/instacart/formula/internal/SynchronizedEventQueue.kt
Laimiux: Also, we are using a non-blocking (atomic) approach to take over processing so the penalty should be relatively small.

formula/src/main/java/com/instacart/formula/internal/SynchronizedUpdateQueue.kt
```diff
@@ -0,0 +1,116 @@
...
+     * thread to continue.
+     */
+    fun postUpdate(update: () -> Unit) {
+        val currentThread = Thread.currentThread()
+        val owner = threadRunning.get()
+        if (owner == currentThread) {
+            // This indicates a nested update where an update triggers another update. Given we
+            // are already thread gated, we can execute this update immediately without a need
+            // for any extra synchronization.
+            update()
```
jasonostrander, line 50: I think we may be missing this case in the unit tests?

jasonostrander (approved):
Looks good to me!

formula/src/main/java/com/instacart/formula/internal/SynchronizedUpdateQueue.kt
```diff
@@ -0,0 +1,116 @@
...
+            return
+        }
+
+        val success = if (updateQueue.peek() == null) {
+            // No pending update, let's try to run our update immediately
+            takeOver(currentThread) { update() }
```
alexanderbezverhni, line 56: nit: `takeOver(currentThread) { update() }` -> `takeOver(currentThread, update)` to avoid anonymous lambda?

formula/src/main/java/com/instacart/formula/internal/SynchronizedUpdateQueue.kt
```diff
@@ -0,0 +1,116 @@
...
+        val owner = threadRunning.get()
+        if (owner == currentThread) {
+            // This indicates a nested update where an update triggers another update. Given we
+            // are already thread gated, we can execute this update immediately without a need
+            // for any extra synchronization.
+            update()
+            return
+        }
+
+        val success = if (updateQueue.peek() == null) {
```
alexanderbezverhni, line 54: `success` -> `updateExecuted`?

formula/src/main/java/com/instacart/formula/internal/SynchronizedUpdateQueue.kt
```diff
@@ -0,0 +1,116 @@
...
+     * Tries to drain the update queue. It will process one update at a time until
+     * queue is empty or another thread takes over processing.
+     */
+    private fun tryToDrainQueue(currentThread: Thread) {
+        while (true) {
+            // First, we peek to see if there is a value to process.
+            val peekUpdate = updateQueue.peek()
+            if (peekUpdate != null) {
+                // Since there is a pending update, we try to process it.
+                val success = takeOver(currentThread) {
```
alexanderbezverhni, line 77: same here - `success` might have a more readable name

formula/src/main/java/com/instacart/formula/internal/ThreadChecker.kt
```diff
@@ -1,19 +0,0 @@
-package com.instacart.formula.internal
-
-import com.instacart.formula.IFormula
-
-/**
- * A poor man's thread checker.
```
alexanderbezverhni, line 6: 😆

formula/src/test/java/com/instacart/formula/subjects/MultiThreadRobot.kt
```diff
@@ -0,0 +1,87 @@
...
+
+        executor.execute {
+            observer.output {
+                function()
+                completionLatch.countDown()
+            }
+        }
+    }
+
+    fun input(newKey: String) = apply {
```
alexanderbezverhni, line 53: This comment is not related to this line. So we're handing off Formula update events to the currently processing Thread, which might be a network/IO thread. I'm not convinced this is the most effective use of our resources. Background network thread is designed to take care of network calls boilerplate, it's not conceptually designed to process app UI state management updates. I think the further network threads usually go is up to parsing response to a more digestible data. 

I'm curious if we should explore Formula runtime having 1 working background thread that runs transitions and side effects. If a side effect needs to run on a main thread, there would be an explicit API to do so, such as:
```
transition(state = state.copy(...)) {
    uiThread {
        ...    
    }
    backgroundThread {
        ...    
    }
}
```

formula/src/main/java/com/instacart/formula/internal/SynchronizedUpdateQueue.kt
ntoskrnl: could be impossible because `takeOver` is inline

formula/src/test/java/com/instacart/formula/subjects/MultiThreadRobot.kt
ntoskrnl: Good point about IO-thread. So, we basically have 2 options: 
- don't perform any enforcement and let user switch thread manually via in the `RxAction.fromObservable{}` block via `observeOn(computation())`.
- or maybe let formula enforce thread. E.g. we can introduce background thread (or a pool of threads) and if the event received not on UI thread and not on one of our background threads, we execute it on the background thread from the pool.

I would prefer the second option.

The API to request on which thread perform side effect is useful. I only see one downside: it could add mental overhead to the formula user and code-reviewer. Engineers will need to spend additional time to decide/verify what should be on `uiThread` what should be on `backgroundThread`. I can be useful though.

Laimiux: I might be wrong, but I’m not sure the statement that “network/IO thread is not conceptually designed to process app UI state management updates” is correct. The thread classification such as I/O or computation only matters in terms of thread pools (and how many threads are allocated). I’m not sure there is a problem that a network thread executes more code (and will exit shortly after). Not having to jump a thread to process the update has two benefits:
- We process the formula update much faster (jumping to the main thread can have a huge 100+ms delay; I’ve seen 300+ms during app start)
- We potentially defer some processing to a background thread, leaving the main thread more responsive

> I'm curious if we should explore Formula runtime having 1 working background thread

The problem with using a single background worker is that each user event will have to jump from main thread to background thread and then back to the main thread. The delay added by this can make things feel less responsive. This is why I think better option is not to confine the formula to a single thread which would require constant jumping back and forth. The current approach tries to execute things on the thread that has the event.

> If a side effect needs to run on a main thread, there would be an explicit API to do so, such as:

In terms of extra threading APIs for side-effect execution, it might be convenient, but I'm not sure it's the right approach. Currently, the formula doesn't manage any threads and doesn't know about the main thread. Not sure introducing those for this use-case is worth it. Regardless if we introduce this threading API within side-effects, the end user will still have to think much more carefully about shared mutable state and how to synchronize reads/writes. I suspect logic to ensure thread-safety should live within classes that we call the side-effects on instead of formula.

formula/src/main/java/com/instacart/formula/internal/SynchronizedEventQueue.kt
tripathiamritansh: ok that makes sense

tripathiamritansh (approved):

formula/src/main/java/com/instacart/formula/internal/SynchronizedUpdateQueue.kt
Laimiux: Removed inline function and updated to `takeOver(currentThread, update)`

formula/src/main/java/com/instacart/formula/internal/Frame.kt
```diff
@@ -14,7 +14,7 @@ import com.instacart.formula.Evaluation
  */
 internal class Frame<Input, State, Output>(
     val input: Input,
-    val state: State,
```
Laimiux, line 17: Unrelated, making private since it's not used

formula/src/main/java/com/instacart/formula/internal/ActionKey.kt
```diff
@@ -2,5 +2,5 @@ package com.instacart.formula.internal
 
 internal data class ActionKey(
     val id: Long,
-    val delegateKey: Any?,
```
Laimiux, line 5: Unrelated, making private since it's not used

alexanderbezverhni (approved):

PR #337, opened 2024-02-13
New Effect type to enable thread specification.

Laimiux:
## What

In the current design, the side effects are not explicitly confined to a specific thread. This can create a problem when we are trying to incrementally adopt multi-threaded formula evaluations. To make things easier to reason about, I will add `DispatcherType` to each side-effect block and default it to the main thread. This will ensure that the default behavior is more predictable and the user of API can opt into unconfined side-effect behavior. The API will look like
```kotlin
transition(Effect.Unconfined) {
    // Unbound side-effect, should only call thread-safe functions
}

transition(newState, Effect.Unconfined) {
  // Unbound side-effect, should only call thread-safe functions
}
```

Since Formula doesn't know about the main thread, we will add public API via the plugin mechanism to specify. 
```kotlin
class MainDispatcher : Dispatcher {}


class MyPlugin : Plugin() {
    override fun mainThreadDispatcher(): Dispatcher {
        return MainDispatcher()
    }
}
```


formula/src/main/java/com/instacart/formula/DispatcherType.kt
```diff
@@ -0,0 +1,24 @@
+package com.instacart.formula
+
+/**
+ * Specifies which thread should execute the side-effect.
+ *
+ * TODO: is the name right?
+ */
+enum class DispatcherType {
```
Laimiux, line 8: I'm still debating the naming here

formula/src/main/java/com/instacart/formula/DispatcherType.kt
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Any thread indicates that side-effect will be executed on
+     * the thread that event arrived on.
+     */
+    ANY,
```
Laimiux, line 13: Still thinking if `ANY` is better than `UNCONFINED`

formula/src/main/java/com/instacart/formula/Effect.kt
```diff
@@ -1,13 +1,16 @@
...
  * [FormulaRuntime] usually after [Formula] state changes are applied. This function is
  * the place to call listeners, log events, trigger network requests or database writes,
  * and etc. Instead of executing effects within [Transition.toResult], deferring execution
  * allows us to ensure that [Formula] is always in the correct state in case effects
  * trigger a new event.
  */
-fun interface Effects {
-    fun execute()
+data class Effect(
+    val dispatcherType: DispatcherType,
```
Laimiux, line 12: Each effect has a `DispatcherType`

formula/src/main/java/com/instacart/formula/Transition.kt
```diff
@@ -39,24 +42,24 @@ fun interface Transition<in Input, State, in Event> {
...
         /**
          * Effects is an optional function returned by [toResult] which will be executed by
          * [FormulaRuntime]. This function is the place to call listeners, log events, trigger
          * network requests or database writes, and etc.
          *
-         * @see Effects
+         * @see Effect
          */
-        abstract val effects: Effects?
+        abstract val effects: List<Effect>
```
Laimiux, line 62: Needs to be a list of effects where each has a dispatcher type.

carrotkite:
## Danger has errored
[!] Invalid `Dangerfile` file: No coverage data found for com/instacart/formula/internal/ManagerDelegate. Updating the Danger gem might fix the issue. Your Danger version: 8.6.1, latest Danger version: 9.4.3

```
 #  from Dangerfile:3
 #  -------------------------------------------
 #  jacoco.minimum_class_coverage_percentage = 0
 >  jacoco.report "/home/runner/jacoco/jacoco.xml"
 #  
 #  -------------------------------------------
```

</p>

formula/src/main/java/com/instacart/formula/Effect.kt
```diff
@@ -0,0 +1,47 @@
+package com.instacart.formula
+
+/**
+ * Effect is a function returned within [Transition.Result] which will be executed by
+ * [FormulaRuntime] usually after [Formula] state changes are applied. This function is
```
alexanderbezverhni, line 5: I think `usually` might be misleading here. Is your intention that usually new state is provided, and effects are *always* executed after new state application? It might be misread as "sometimes effects are executed before state changes application".

formula/src/main/java/com/instacart/formula/Effect.kt
```diff
@@ -0,0 +1,47 @@
...
+     * An effect type which defines the execution timing and threading of the effect. Take
+     * a look at [Unconfined], [Main], [Background].
+     */
+    sealed class Type
+
+    /**
+     * Unconfined effect type indicates that the effect can run on any thread and will be
+     * executed after all state changes are processed.
+     */
+    data object Unconfined : Type()
```
alexanderbezverhni, line 27: nit: should all implementing classes live inside `Type`?

formula/src/main/java/com/instacart/formula/Effect.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Immediate effect type indicates that the effect can run on any thread and can be executed
+     * immediately before the state changes are processed and formula is re-evaluated. This
+     * is a useful optimization for inlining multiple state changes as part of a single
+     * re-evaluation.
```
alexanderbezverhni, line 33: Not sure I'm following here. Isn't it in Formula's contract that effects are guaranteed to be executed *after* state changes application?

formula/src/main/java/com/instacart/formula/Effect.kt
```diff
@@ -0,0 +1,47 @@
...
+     */
+    // TODO: this will be added in a future.
+//    data object Immediate: Type()
+
+    /**
+     * Will schedule side-effect to be executed on the main thread.
```
alexanderbezverhni, line 39: Not directly related to this PR, b/c we already started to articulate "main thread" terminology in other places - should this "main-thread" limitations be exclusive to `formula-android` only? If we claim `formula` to be Android-agnostic and JVM-friendly or even kotlin-multiplatform-friendly, the concept of "only main thread is allowed to touch UI" is irrelevant there.

formula/src/main/java/com/instacart/formula/TransitionContext.kt
```diff
@@ -21,23 +21,56 @@ interface TransitionContext<out Input, State> {
...
+    fun <State> transition(
+        state: State,
+        effect: () -> Unit,
+    ): Transition.Result.Stateful<State> {
+        return transition(state, Effect.Main, effect)
+    }
+
+    /**
+     * Returns a result that contains a new [State] object and optional effect
+     * executable that will be executed after the state is updated.
```
alexanderbezverhni, line 45: This contradicts with aforementioned.

formula/src/main/java/com/instacart/formula/Effect.kt
Laimiux: I think that the public API is better as such (as type is implied by `Effect.`) 
```kotlin
transition(Effect.Unconfined) { }
```
rather than
```kotlin
transition(Effect.Type.Unconfined) {}
```

formula/src/main/java/com/instacart/formula/Effect.kt
Laimiux: It's for special cases where you might not want that behavior. For example, if a child formula updates its state and then notifies the parent via callback to also update. Currently, this would have two full paths to evaluate child state change, perform the side-effect, and then evaluate the parent state. With the `Immediate` type, we would update child state, then the parent state, and only then, do a single re-evaluation.

Laimiux: Note: this will be implemented in the future (just added placeholder for the type to communicate the purpose)

formula/src/main/java/com/instacart/formula/Effect.kt
Laimiux: `Effect.Main` and `Effect.Background` are metadata for `Plugin` to decide how to interpret. If a platform doesn't support `Main`, it doesn't have to do anything. I could have named it `Effect.Confined`, but thought `Effect.Main` seems clearer though less generic.

formula/src/main/java/com/instacart/formula/Effect.kt
Laimiux: Good point, updated

formula/src/main/java/com/instacart/formula/TransitionContext.kt
Laimiux: Updated documentation

alexanderbezverhni (approved):

PR #360, opened 2024-03-18
Reset StateBatchScheduler state on error.

Laimiux:


formula/src/main/java/com/instacart/formula/batch/StateBatchScheduler.kt
```diff
@@ -45,18 +45,21 @@ class StateBatchScheduler : BatchScheduler {
...
+            update()
+        } finally {
+            isUpdating.remove()
 
-        // Get collected batches
-        val batches: MutableSet<BatchScheduler.Batch>? = threadLocalBatches.get()
-        // Clear them from our state
-        threadLocalBatches.remove()
+            // Get collected batches
+            val batches: MutableSet<BatchScheduler.Batch>? = threadLocalBatches.get()
```
Laimiux, line 54: We clear try to close out open batches even if there was an error

jasonostrander (approved):

PR #364, opened 2024-04-03
Handling errors from child formulas

jhowens89:


carrotkite:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">No coverage data found for com/instacart/formula/internal/FormulaManager</td>
    </tr>
  </tbody>
</table>

### JaCoCo Code Coverage 83.83% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/internal/FormulaManager` | No coverage data found : -% | No coverage data found : -% | :black_joker: |
| `com/instacart/formula/internal/SnapshotImpl` | 87% | 0% | :white_check_mark: |
| `com/instacart/formula/internal/FormulaManagerImpl` | 85% | 0% | :white_check_mark: |
| `com/instacart/formula/FormulaContext` | 86% | 0% | :white_check_mark: |


</p>

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -245,6 +245,40 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
+    ): ChildOutput? {
+        val childrenManager = getOrInitChildrenManager()
+        val manager = childrenManager.findOrInitChild(key, formula, input)
+
+        // If termination happens while running, we might still be initializing child formulas. To
+        // ensure correct behavior, we mark each requested child manager as terminated to avoid
+        // starting new actions.
+        if (isTerminated()) {
+            manager.markAsTerminated()
+            manager.performTerminationSideEffects()
```
jhowens89, line 264: I'm concerned that performTerminationSideEffects will get executed more than necessary. I've added some code to poll the transition queue on termination to avoid duplicate effect execution, but calling this more than once will have undesired effects such as an extra inspector::onFormulaFInished call. 

@Laimiux can you elaborate a little on the expectations of when performTerminationSideEffects is going to be called?

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -245,6 +245,40 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
+        }
+        manager.setValidationRun(isValidationEnabled)
+
+        return try {
+            if (!manager.isTerminated()) {
```
jhowens89, line 269: The preceding performTerminationSideEffects was added because we are not executing the child formula in the this condition. My reasoning is that there is a good chance this will throw another exception and continually throwing every evaluation is a performance concern.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -130,28 +130,30 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
-            val evaluationId = globalEvaluationId
-            val evaluation = if (firstRun) {
-                firstRun = false
-                evaluation(input, evaluationId)
-            } else if (lastFrame == null || isEvaluationNeeded(lastFrame.associatedEvaluationId)) {
-                evaluation(input, evaluationId)
-            } else {
-                lastFrame.evaluation
-            }
+        try {
```
jhowens89, line 133: Wrapped this is a try/finally to keep isRunning in a good state. Not sure if it is completely necessary

formula/src/main/java/com/instacart/formula/FormulaContext.kt
```diff
@@ -102,6 +102,18 @@ abstract class FormulaContext<out Input, State> internal constructor(
         input: ChildInput
     ): ChildOutput
 
+    /**
+     * Returns the latest output of the [child] formula or null if the child formula has encountered
+     * an exception. Child formulas that encounter exceptions will be be terminated and will not
+     * be run again. Formula runtime ensures the [child] is running, manages its internal state
+     * and will trigger `evaluate` if needed.
+     */
+    abstract fun <ChildInput, ChildOutput> child(
```
jhowens89, line 111: Use separate method in FormulaContext to avoid nullability concerns complicating API contract of FormulaManager

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -245,6 +247,40 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
+                manager.run(input).output
+            } else {
+                null
+            }
+        } catch (e: ValidationException) {
+            throw e
+        } catch (e: Throwable) {
+            manager.markAsTerminated()
+            onError(e)
+            manager.performTerminationSideEffects()
```
jhowens89, line 279: performanceTerminationSideEffects could throw if the problematic code is within a transition. Not sure if that's something we should guard against.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -245,6 +247,40 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
+        input: ChildInput,
+        onError: (Throwable) -> Unit,
+    ): ChildOutput? {
+        val childrenManager = getOrInitChildrenManager()
+        val manager = childrenManager.findOrInitChild(key, formula, input)
+
+        // If termination happens while running, we might still be initializing child formulas. To
+        // ensure correct behavior, we mark each requested child manager as terminated to avoid
+        // starting new actions.
+        if (isTerminated()) {
```
Laimiux, line 262: I'm not sure how you would hit this

jhowens89: I'm not entirely sure either, but it copied from the code here: https://github.com/instacart/formula/blob/15351c18b75be344622b9785649e92efb1778a15/formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt#L243

formula/src/main/java/com/instacart/formula/FormulaContext.kt
Laimiux: There is also a method without an `input` property for `IFormula<Unit, T>`, we should similarly add `onError` support.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
Laimiux: Oh, I think it's if a side-effect terminates the root formula.

formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
Laimiux: This is an excellent question that is a bit broader than this specific pull-request:
-  Currently, if `performTerminationSideEffects` crashes, we can leak memory by leaving an action(s) still running. 
- Related to this pull-request, if formula crashed, do we need to execute its remaining transitions? My take would be no

formula/src/main/java/com/instacart/formula/internal/FormulaManager.kt
```diff
@@ -27,6 +29,11 @@ interface FormulaManager<Input, Output> {
 
     /**
      * Called when we are ready to perform termination side-effects.
+     * @param executeTransitionQueue whether the formula should execute remaining transitions
+     *      that accumulated while it was running. False is passed in the case where the
+     *      formula threw an exception and we want to avoid executubg transitions that have
```
Laimiux, line 34: Nit: `executubg` typo

Laimiux (approved):

PR #366, opened 2024-04-24
Ensure event ordering with dispatching.

Laimiux:
### The problem
With dispatching, we introduced the possibility for events to arrive in an unexpected order. For example:
1. We start a formula on a background thread 
2. During the start, formula subscribes to a relay that emits an initial value 
3. This initial value needs to be dispatched to the main thread
4. Before dispatching completes, the relay emits another value already on the main thread

The following scenario leads us to process the new value before the initial value. So, if the new value is a content event and the initial value is a loading event, you will first set the state to content and then to loading which will lead to infinite loading bug in your feature. 

### The fix
To fix this problem, we need to keep a predictable order of the incoming events. Instead of dispatching `queue.postUpdate`, we will pass `dispatcher` to `SynchronizedUpdateQueue`. `SynchronizedUpdateQueue` will add the update to internal queue to keep event order and will dispatch request to process the queue. This way we maintain the update order while processing can happen on a different thread.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
@@ -94,7 +94,9 @@ class FormulaRuntime<Input : Any, Output : Any>(
     }
 
     fun onInput(input: Input) {
-        synchronizedUpdateQueue.postUpdate { onInputInternal(input) }
+        defaultDispatcher.dispatch {
+            synchronizedUpdateQueue.postUpdate { onInputInternal(input) }
+        }
     }
 
     private fun onInputInternal(input: Input) {
--- a/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
@@ -94,9 +94,7 @@ class FormulaRuntime<Input : Any, Output : Any>(
     }
 
     fun onInput(input: Input) {
-        defaultDispatcher.dispatch {
-            synchronizedUpdateQueue.postUpdate { onInputInternal(input) }
-        }
+        synchronizedUpdateQueue.postUpdate(defaultDispatcher) { onInputInternal(input) }
     }
 
     private fun onInputInternal(input: Input) {
@@ -137,7 +135,7 @@ class FormulaRuntime<Input : Any, Output : Any>(
     }
 
     fun terminate() {
-        synchronizedUpdateQueue.postUpdate(this::terminateInternal)
+        synchronizedUpdateQueue.postUpdate(defaultDispatcher, this::terminateInternal)
     }
 
     private fun terminateInternal() {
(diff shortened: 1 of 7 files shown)


carrotkite:
### JaCoCo Code Coverage 84.38% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/FormulaRuntime` | 94% | 0% | :white_check_mark: |


</p>

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -94,7 +94,7 @@ class FormulaRuntime<Input : Any, Output : Any>(
     }
 
     fun onInput(input: Input) {
-        synchronizedUpdateQueue.postUpdate { onInputInternal(input) }
+        synchronizedUpdateQueue.postUpdate(defaultDispatcher) { onInputInternal(input) }
```
Laimiux, line 97: Fixing an issue where input events didn't use dispatcher

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -135,7 +135,7 @@ class FormulaRuntime<Input : Any, Output : Any>(
     }
 
     fun terminate() {
-        synchronizedUpdateQueue.postUpdate(this::terminateInternal)
+        synchronizedUpdateQueue.postUpdate(defaultDispatcher, this::terminateInternal)
```
Laimiux, line 138: Terminate also needs to use a dispatcher

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -174,26 +174,24 @@ class FormulaRuntime<Input : Any, Output : Any>(
 
     override fun executeBatch(updates: List<() -> Unit>) {
         // Using default dispatcher for batch execution
-        defaultDispatcher.dispatch {
```
Laimiux, line 177: Instead of dispatching and then posting update, we add update to the queue and dispatch request to process. This way the queue keeps track of all events in arrival order while processing can be shifted to a different thread.

formula/src/main/java/com/instacart/formula/plugin/Dispatcher.kt
```diff
@@ -30,10 +39,20 @@ interface Dispatcher {
...
 
     /**
      * Dispatches [executable] to a thread specified by the [Dispatcher].
      */
     fun dispatch(executable: () -> Unit)
+
+    /**
+     * Returns true if dispatching event is needed
+     */
+    fun isDispatchNeeded(): Boolean
```
Laimiux, line 57: Returns true if we need to dispatcher. For example, the main thread dispatcher can check if are already on main thread.

formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
```diff
@@ -660,6 +660,8 @@ class FormulaRuntimeTest(val runtime: TestableRuntime, val name: String) {
             }
     }
 
+    // TODO: does this
+    @Ignore
```
Laimiux, line 664: This behavior changes because all events are added to the queue and processed in order. So `increment` will be processed before `toggleCallback`.

jasonostrander (approved):
LGTM

PR #371, opened 2024-07-15
[Testing] Add ability to check how many child formulas are running.

Laimiux:


carrotkite:
### JaCoCo Code Coverage 83.92% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/test/TestFormula` | 0% | 0% | :white_check_mark: |


</p>

formula-test/src/main/java/com/instacart/formula/test/TestFormula.kt
```diff
@@ -106,4 +106,11 @@ abstract class TestFormula<Input, Output> :
         }
         instance.input.interact()
     }
+
+    fun assertRunningCount(expected: Int) {
+        val count = stateMap.size
+        if (count != expected) {
+            throw AssertionError("Expected $expected running formulas, but there were $count instead")
```
seeming-amusing, line 113: Just a (possible) clarification point, but it might be worth specifying the formula type in this error message (e.g., "Expected $expected ${instance.javaClass.simpleName} instances running, ..."). That way when the error shows up, it is more immediately obvious which one to look for.

seeming-amusing (approved):
LGTM.

formula-test/src/main/java/com/instacart/formula/test/TestFormula.kt
Laimiux: Unfortunately, that requires bigger change as `instance.javaclass.simpleName` won't point to anything useful

PR #375, opened 2024-08-15
Fix a termination race-condition within formula test.

Laimiux:
## What
Fixing a bug in the formula test module where:
- Formula with "A" key is run
- Key A is terminated but side-effect is not executed yet
- New Key A is run which overwrites the previous Key A
- Termination side-effect occurs and removes new Key A 

To fix this problem, I've made a couple of changes
- Added a unique long identifier to ensure that termination only removes the appropriate key
- Moved the side-effect to be executed immediately.

### DIFF
--- a/formula-test/src/main/java/com/instacart/formula/test/TestFormula.kt
+++ b/formula-test/src/main/java/com/instacart/formula/test/TestFormula.kt
@@ -4,7 +4,7 @@ import com.instacart.formula.Action
 import com.instacart.formula.Evaluation
 import com.instacart.formula.Formula
 import com.instacart.formula.Snapshot
-import java.lang.IllegalStateException
+import java.util.concurrent.atomic.AtomicLong
 
 /**
  * Test formula is used to provide a fake formula implementation. It allows you to [send][output]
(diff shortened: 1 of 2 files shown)


formula-test/src/main/java/com/instacart/formula/test/TestFormula.kt
```diff
@@ -62,9 +71,8 @@ abstract class TestFormula<Input, Output> :
             output = state.output,
             actions = context.actions {
                 Action.onTerminate().onEvent {
-                    transition {
-                        stateMap.remove(state.key)
-                    }
+                    stateMap.remove(state.uniqueIdentifier)
```
Laimiux, line 74: Moving the side-effect here so it happens immediately and not after all the state changes are processed.

formula-test/src/main/java/com/instacart/formula/test/TestFormula.kt
```diff
@@ -47,11 +51,16 @@ abstract class TestFormula<Input, Output> :
...
+        return State(
+            uniqueIdentifier = identifierGenerator.getAndIncrement(),
+            key = key(input),
+            output = initialOutput(),
+        )
     }
 
     override fun Snapshot<Input, State<Output>>.evaluate(): Evaluation<Output> {
-        stateMap[state.key] = Value(
+        stateMap[state.uniqueIdentifier] = Value(
```
Laimiux, line 62: Using a unique identifier to avoid collisions of the same key.

carrotkite:
### JaCoCo Code Coverage 83.55% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/test/TestFormula` | 0% | 0% | :white_check_mark: |


</p>

alexanderbezverhni (approved):
Looks good! Is there no coverage for `TestFormula`?

Laimiux:
> Looks good! Is there no coverage for `TestFormula`?

Looks like the tests were not running. I fixed it in this PR, but there seems to be only 14% coverage. I will add more tests in a follow-up

PR #387, opened 2024-09-10
Increase test module coverage.

Laimiux:


formula-test/src/test/java/com/instacart/formula/test/TestActionObserverTest.kt
```diff
@@ -0,0 +1,61 @@
...
+        assertThat(result.exceptionOrNull()).isInstanceOf(AssertionError::class.java)
+    }
+
+    @Test fun values() {
+        val values = multipleValueStream().test().values()
+        assertThat(values).hasSize(2)
+    }
+
+    @Test fun `cancel throws exception if action does not provide cancelable`() {
+        val result = kotlin.runCatching { multipleValueStream().test().cancel() }
```
deesonpatel, line 32: Is there a difference between above `runCatching` block vs `kotlin.runCatching`?

formula-test/src/test/java/com/instacart/formula/test/TestActionObserverTest.kt
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            override fun key(): Any? = null
+        }
+
+        action.test().cancel()
+        assertThat(cancelableCalled).isEqualTo(1)
+    }
+
+    fun multipleValueStream() = object : Action<Int> {
```
deesonpatel, line 52: nit: could be private

deesonpatel (approved):

formula-test/src/test/java/com/instacart/formula/test/TestActionObserverTest.kt
Laimiux: It's the same function, the auto-complete keeps giving me `kotlin.runCatching` instead of `runCatching`

deesonpatel: Oh! That's so weird

PR #398, opened 2024-09-25
Refactor FeatureFactory public API using Params.

Laimiux:
## What
I'm wrapping the `dependencies` and `key` into a `Params` object and updating `FeatureFactory.initialize` to get it via the extension function receiver type `Params.initialize(): Feature`.

Feature factory initialization changes from
```kotlin
override fun initialize(dependencies: Unit, key: Key): Feature
```
to 
```kotlin
override fun Params.initialize(): Feature
```

### Why
We need to pass `FragmentId` or something equivalent to the `FeatureFactory`. This change requires us to introduce a new parameter that would break existing usages. Instead of just adding a new parameter, I'm refactoring the way this public API works which will enable us to modify `Params` object in the future without breaking the usages.

### DIFF
--- a/formula-android/src/main/java/com/instacart/formula/android/FeatureFactory.kt
+++ b/formula-android/src/main/java/com/instacart/formula/android/FeatureFactory.kt
@@ -35,10 +35,22 @@ package com.instacart.formula.android
  * @param Key a type of fragment key that is used to identify this feature.
  *
  */
-interface FeatureFactory<in Dependencies, in Key : FragmentKey> {
+abstract class FeatureFactory<in Dependencies, in Key : FragmentKey> {
+
+    inner class Params(
+        val dependencies: @UnsafeVariance Dependencies,
+        val key: @UnsafeVariance Key,
+    )
+
+    /**
+     * Initializes the [Feature] using [Params] provided.
+     */
+    abstract fun Params.initialize(): Feature
 
     /**
      * Initializes state observable and a view factory for a specific [key].
      */
-    fun initialize(dependencies: Dependencies, key: Key): Feature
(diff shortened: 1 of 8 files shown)


carrotkite:
### JaCoCo Code Coverage 97.15% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/android/FeatureFactory` | 100% | 0% | :white_check_mark: |
| `com/instacart/formula/android/internal/MappedFeatureFactory` | 100% | 0% | :white_check_mark: |


</p>

formula-android/src/main/java/com/instacart/formula/android/FeatureFactory.kt
```diff
@@ -35,10 +35,22 @@ package com.instacart.formula.android
...
+     */
+    abstract fun Params.initialize(): Feature
 
     /**
      * Initializes state observable and a view factory for a specific [key].
      */
-    fun initialize(dependencies: Dependencies, key: Key): Feature
+    fun initialize(dependencies: Dependencies, key: Key): Feature {
+        return Params(dependencies, key).initialize()
+    }
```
alexanderbezverhni, line 55: Will this become `private` in a follow up?

alexanderbezverhni (approved):

formula-android/src/main/java/com/instacart/formula/android/FeatureFactory.kt
Laimiux: Probably not, this function provides a better API to initialize the `Feature`. It's a final function that delegates to a better internal API for implementation. 
- for implentee use `Params.initialize` 
- for the consumer use `initialize(dependencies, key, other params)`

PR #414, opened 2025-04-03
Ensure inspector.onRunFinished() is called.

Laimiux:
Small change to ensure `inspector.onRunFinished` is called even if run formula errors out.


setheclark (approved):
Looks good.  Should we add testing for this?

Laimiux:
> Looks good. Should we add testing for this?

In some way, it's covered by existing tests

PR #416, opened 2025-05-06
Add CoroutineAction to support suspend functions.

Laimiux:
## What
Adding support for launching coroutines as part of formula actions.

### DIFF
--- a/formula-coroutines/src/main/java/com/instacart/formula/coroutines/CoroutineAction.kt
+++ b/formula-coroutines/src/main/java/com/instacart/formula/coroutines/CoroutineAction.kt
@@ -1,0 +1,106 @@
+package com.instacart.formula.coroutines
+
+import com.instacart.formula.Action
+import com.instacart.formula.Cancelable
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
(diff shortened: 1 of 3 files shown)


formula-coroutines/src/main/java/com/instacart/formula/coroutines/CoroutineAction.kt
```diff
@@ -0,0 +1,106 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+/**
+ *
+ * Adapter which allows to run a coroutine as an [Action].
+ */
+class CoroutineAction<Result> @PublishedApi internal constructor(
```
Laimiux, line 17: Given that coroutines are the default technology, this will likely be moved to `formula` core module in the future and might be merged with `Action` type itself.

formula-coroutines/src/test/java/com/instacart/formula/coroutines/CoroutineActionTest.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    @Test fun `launch - dispatcher`() {
+        val dispatcher = StandardTestDispatcher()
+        val action = CoroutineAction.launch(coroutineDispatcher = dispatcher) {
+            delay(1000)
+            "Result"
+        }
+
+        val observer = action.test()
+        dispatcher.scheduler.advanceUntilIdle()
```
jasonostrander, line 36: Should maybe assert no values until you advance the dispatcher?

jasonostrander (approved):
Looks good!

PR #444, opened 2025-06-30
[Android] Move FragmentEnvironment into FragmentStore.

Laimiux:
## What
Moving `FragmentEnvironment` to `FragmentStore`. It doesn't make sense to pass `FragmentEnvironment` if there is no `FragmentStore`


formula-android/src/main/java/com/instacart/formula/android/internal/ActivityManager.kt
```diff
@@ -103,7 +100,7 @@ internal class ActivityManager<Activity : FragmentActivity>(
     private fun subscribeToFragmentStateChanges(): Disposable {
         return store
             .fragmentStore
-            .state(environment)
```
Laimiux, line 106: No need to pass it manually anymore.

PR #451, opened 2025-07-21
Catch exceptions during action initialization, action termination, and side-effects.

Laimiux:
## What
Instead of crashing the whole formula tree, we are isolating action errors.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/Effect.kt
+++ b/formula/src/main/java/com/instacart/formula/Effect.kt
@@ -1,5 +1,6 @@
 package com.instacart.formula
 
+import com.instacart.formula.plugin.FormulaError
 import com.instacart.formula.plugin.Plugin
 
 /**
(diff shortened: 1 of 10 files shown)


formula/src/main/java/com/instacart/formula/DeferredAction.kt
```diff
@@ -28,11 +28,23 @@ class DeferredAction<Event>(
             }
         }
 
-        cancelable = action.start(scope, emitter)
+        try {
+            cancelable = action.start(scope, emitter)
```
Laimiux, line 32: Catch exceptions during action initialization

formula/src/main/java/com/instacart/formula/DeferredAction.kt
```diff
@@ -28,11 +28,23 @@ class DeferredAction<Event>(
...
+
     }
 
-    internal fun tearDown() {
-        cancelable?.cancel()
+    internal fun tearDown(formulaType: Class<*>) {
+        try {
+            cancelable?.cancel()
+        } catch (throwable: Throwable) {
+            val error = FormulaError.ActionError(formulaType, throwable)
```
Laimiux, line 44: Catch exeptions during action termination

jasonostrander (approved):
Looks good. I assume these errors get logged to sentry as errors? My only concern here is that the previous fail-fast behavior would be obvious, and this new behavior will likely result in some non-obvious downstream issues

Laimiux:
> Looks good. I assume these errors get logged to sentry as errors? My only concern here is that the previous fail-fast behavior would be obvious, and this new behavior will likely result in some non-obvious downstream issues

Yeah, we will have global logging for these errors. Currently, an exception leads to a mostly non-functional UI due to a terminated formula. Given it doesn't crash the app, I wouldn't call the current behavior obvious. There could be some non-obvious behavior when a certain part of the system fails, but given the declarative nature of Formula, the surface area should be quite small.

PR #453, opened 2025-07-22
Emit errors only via global handler.

Laimiux:
## What
Instead of emitting formula errors via `RxJava` or Coroutines, we emit them via a consistent formula error mechanism.


formula-rxjava3/src/main/java/com/instacart/formula/rxjava3/RxJavaRuntimeExtensions.kt
```diff
@@ -40,7 +40,6 @@ fun <Input : Any, Output : Any> start(
             config = config ?: RuntimeConfig(),
         )
         runtime.setOnOutput(emitter::onNext)
-        runtime.setOnError(emitter::onError)
```
Laimiux, line 43: There is no need to emit the error as we rely on the global error listener instead.

formula/src/main/java/com/instacart/formula/RuntimeExtensions.kt
```diff
@@ -43,10 +43,6 @@ fun <Input : Any, Output : Any> IFormula<Input, Output>.runAsStateFlow(
     // having a specific dispatcher does not make sense.
     val config = RuntimeConfig(defaultDispatcher = Dispatcher.None)
     val runtime = asRuntime(scope.coroutineContext, config)
-    runtime.setOnError { e ->
```
Laimiux, line 46: This logic has been moved to `FormulaRuntime`

formula/src/main/java/com/instacart/formula/RuntimeExtensions.kt
```diff
@@ -91,7 +87,6 @@ private fun <Input : Any, Output : Any> start(
     val callbackFlow = callbackFlow {
         val runtime = formula.asRuntime(coroutineContext, config)
         runtime.setOnOutput(this::trySend)
-        runtime.setOnError(this::close)
```
Laimiux, line 94: There is no need to emit the error as we rely on the global error listener instead.

PR #454, opened 2025-07-22
Use onError to emit duplicate child key errors.

Laimiux:
## What
Use `onError` for duplicate child key errors instead of a custom function.


formula/src/main/java/com/instacart/formula/FormulaPlugins.kt
```diff
@@ -30,14 +30,6 @@ object FormulaPlugins {
         plugin?.onError(error)
     }
 
-    fun onDuplicateChildKey(
```
Laimiux, line 33: Removing this in favor of using `FormulaPlugins.onError` with `FormulaError`

formula/src/test/java/com/instacart/formula/FormulaRuntimeTest.kt
```diff
@@ -814,7 +815,8 @@ class FormulaRuntimeTest {
             }
         }
 
-        parentFormula.test().input(Unit).output {
+        // Need failOnError to avoid test failing due to duplicate child errors
+        parentFormula.test(failOnError = false).input(Unit).output {
```
Laimiux, line 819: Previously, duplicate child errors were not emitted in tests.

PR #456, opened 2025-07-23
Isolating formula evaluation error handling.

Laimiux:
## What
Instead of always propagating formula errors up to the top (unless `onError` is provided with `child`), we are updating API to be a bit more resilient:
- Automatically log all formula errors with the global error handler
- Replace child `onError` API with `childOrNull` (it will automatically log the error)
- Regular `child` function will return the last computed output value or propagate the error up to the parent formula. In many cases, the failure will be isolated to the first two layers.

### DIFF
--- a/formula/src/main/java/com/instacart/formula/FormulaContext.kt
+++ b/formula/src/main/java/com/instacart/formula/FormulaContext.kt
@@ -94,20 +94,6 @@ abstract class FormulaContext<out Input, State> internal constructor(
     }
 
     /**
-     * A convenience method to run a formula that takes no input. Returns the latest output of
-     * the [child] formula or null if the child formula has encountered an exception. Child formulas
-     * that encounter exceptions will be be terminated and will not be run again. Formula runtime
-     * ensures the [formula] is running, manages its internal state and will trigger `evaluate`
-     * if needed.
-     */
-    fun <ChildOutput> child(
-        formula: IFormula<Unit, ChildOutput>,
-        onError: (Throwable) -> Unit,
-    ): ChildOutput? {
-        return child(formula, Unit, onError)
-    }
-
-    /**
      * Returns the latest output of the [child] formula. Formula runtime ensures the [child]
      * is running, manages its internal state and will trigger `evaluate` if needed.
      */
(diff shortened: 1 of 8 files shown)


formula/src/main/java/com/instacart/formula/internal/FormulaManagerImpl.kt
```diff
@@ -151,7 +151,23 @@ internal class FormulaManagerImpl<Input, State, Output>(
...
+            return result.output
+        } catch (e: Throwable) {
+            if (e is ValidationException) {
+                throw e
+            } else {
+                markAsTerminated()
+                performTerminationSideEffects()
+
+                if (e !is CascadingFormulaException) {
+                    val error = FormulaError.Unhandled(formulaType, e)
```
Laimiux, line 163: To avoid logging same issue multiple times, we check that it's not `CascadingFormulaException`

formula/src/main/java/com/instacart/formula/FormulaRuntime.kt
```diff
@@ -292,8 +293,11 @@ class FormulaRuntime<Input : Any, Output : Any>(
             try {
                 terminateInternal()
             } finally {
-                val error = FormulaError.Unhandled(manager.formulaType, e)
-                emitError(error)
+
+                if (e !is CascadingFormulaException) {
```
Laimiux, line 297: Only log exception if it hasn't been logged already

formula/src/main/java/com/instacart/formula/FormulaContext.kt
```diff
@@ -93,39 +93,37 @@ abstract class FormulaContext<out Input, State> internal constructor(
         return child(formula, Unit)
     }
 
+    /**
+     * Returns the latest output of the [child] formula. Formula runtime ensures the [child]
+     * is running, manages its internal state and will trigger `evaluate` if needed.
+     */
```
alexanderbezverhni, line 99: Should we mention different error handling here, as compared to `childOrNull()`?

alexanderbezverhni (approved):
Nice stability improvement!

PR #461, opened 2025-09-26
coroutineScheduler support within TestFormulaObserver

Jawnnypoo:
## Key problem
If we have something like this in a formula:
```
Action.fromFlow {
    flow {
        delay(delay)
        emit(1)
    }
}.onEvent {
    transition(state.copy(count = state.count + 1))
}
```
then in tests, we don't currently have a way to "advance" time cleanly for the given formula action. 

## Solution
This adds a param to `TestFormulaObserver` so that we can pass in our own `TestCoroutineScheduler`. We can then call APIs such as `advanceTimeBy ` and `advanceUntilIdle` within our unit tests to simulate the passage of time, without having to actually wait the actual time with things like Thread.sleep.

```
val scheduler = TestCoroutineScheduler()
val observer = formula.test(coroutineScheduler = scheduler)
scheduler. advanceUntilIdle()
```
You can see the added `TestCoroutineSchedulerIntegrationTest` for a complete example.

### DIFF
--- a/formula-test/src/test/java/com/instacart/formula/test/TestCoroutineSchedulerIntegrationTest.kt
+++ b/formula-test/src/test/java/com/instacart/formula/test/TestCoroutineSchedulerIntegrationTest.kt
@@ -1,11 +1,6 @@
 package com.instacart.formula.test
 
-import com.instacart.formula.Action
-import com.instacart.formula.Evaluation
-import com.instacart.formula.Formula
-import com.instacart.formula.Snapshot
-import kotlinx.coroutines.delay
-import kotlinx.coroutines.flow.flow
+import com.google.common.truth.Truth
 import kotlinx.coroutines.test.TestCoroutineScheduler
 import kotlinx.coroutines.test.runTest
 import org.junit.Test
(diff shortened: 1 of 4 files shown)


carrotkite:
### JaCoCo Code Coverage 94.13% :white_check_mark:
|:---|:---:|:---:|:---:|
| `com/instacart/formula/test/TestFormulaObserver` | 90% | 0% | :white_check_mark: |


</p>

Jawnnypoo:
I am thinking we might need to think about this more for the case of child formulas being used in tests, since we do not have direct access to give them a test scheduler.

formula-test/src/test/java/com/instacart/formula/test/TestCoroutineSchedulerIntegrationTest.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.coroutines.test.TestCoroutineScheduler
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+import kotlin.time.Duration.Companion.seconds
+
+class TestCoroutineSchedulerIntegrationTest {
+
+    @Test
+    fun `delay within action is advanced by scheduler`() = runTest {
+        val scheduler = TestCoroutineScheduler()
```
Laimiux, line 13: Same here, we should use `scheduler` provided by the test

Laimiux (approved):

